PR #11, opened 2013-07-07
Fix linkedit errors and compiler warnings on OS X

wjwwood:
This fixes several things:

c45b7a5 fixes linkedit errors:

```
Linking CXX shared library /Users/william/nav_ws/devel_isolated/tf2/lib/libtf2.dylib
Undefined symbols for architecture x86_64:
  "console_bridge::log(char const*, int, console_bridge::LogLevel, char const*, ...)", referenced from:
      tf2::BufferCore::warnFrameId(char const*, std::string const&) const in buffer_core.cpp.o
      tf2::BufferCore::setTransform(geometry_msgs::TransformStamped_<std::allocator<void> > const&, std::string const&, bool) in buffer_core.cpp.o
      tf2::BufferCore::lookupTransform(std::string const&, std::string const&, ros::Time const&) const in buffer_core.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

b45a609 fixes compiler warnings:

```
/Users/william/nav_ws/install_isolated/include/tf2/buffer_core.h:90:23: warning: in-class initializer for static data member of type 'const double' is a GNU extension [-Wgnu]
  static const double DEFAULT_CACHE_TIME = 10.0;  //!< The default amount of time to cache data in seconds
```

This also removes an OS X specific fix which no longer applies. It now uses an int, but if that isn't ok we can use a define instead.

c5dbb6a Fixes a compiler warning by replacing the use of pointers with shared_ptr's:

```
/Users/william/nav_ws/src/geometry_experimental/tf2/src/buffer_core.cpp:273:5: warning: delete called on 'tf2::TimeCacheInterface' that is abstract but has non-virtual destructor
      [-Wdelete-non-virtual-dtor]
    delete frame_ptr;
```

Since it now uses a shared_ptr, there is no need to explicitly delete the object.

106f46d Fixes these compiler warnings:

```
/Users/william/nav_ws/src/tf/src/pytf.cpp:471:16: warning: implicit conversion of NULL constant to 'int' [-Wnull-conversion]
  {NULL, NULL, NULL},
  ~            ^~~~
               0
```

@tfoote thoughts?

### DIFF
--- a/tf2/CMakeLists.txt
+++ b/tf2/CMakeLists.txt
@@ -2,8 +2,8 @@ cmake_minimum_required(VERSION 2.8.3)
 project(tf2)
 
 find_package(console_bridge REQUIRED)
-find_package(catkin REQUIRED COMPONENTS geometry_msgs tf2_msgs rospy rostime)
-find_package(Boost REQUIRED COMPONENTS thread signals)
+find_package(catkin REQUIRED COMPONENTS geometry_msgs rospy rostime tf2_msgs)
+find_package(Boost REQUIRED COMPONENTS signals system thread)
 
 catkin_package(
    INCLUDE_DIRS include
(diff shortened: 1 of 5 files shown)


tfoote:
Generally it looks good.  

I'm not sure that https://github.com/ros/geometry_experimental/commit/b45a609 will have the same behavior.  A Duration doesn't have a single int constructor which is how the variable is used. It probably will get autocast to a double anyway, and thus might work? Apparently this is fixed in C++11 http://stackoverflow.com/questions/13697265/static-const-double-cannot-have-an-in-class-initializer-why-is-it-so but in the mean time we could think about a DEFINE or verify that the single int gets cast correctly.

wjwwood:
The `const int DEFAULT_CACHE_TIME` is always being passed to `ros::Duration` or to `ros::Time::fromSec`, so I think it is ok to let it get implicitly converted to a double there.

PR #31, opened 2013-10-16
Added 6 param transform again

Garfonso:
Yes, using Euler angles is a bad habit. But it is much more convenient if you just need a rotation by 90° somewhere to set it up in Euler angles. So I added the option to supply only the 3 angles.


tfoote:
Thanks, I'll check this shortly and merge it.

PR #47, opened 2013-11-27
Fix const correctness of tf2::Vector3 rotate() method

roehling:
The method does not modify the class thus should be const.
This has already been fixed in Bullet itself.


tfoote:
Looks good. Thanks.

PR #51, opened 2014-01-29
Use ros::MessageEvent's in tf2_ros for connection information

kwatts:
Patch to stop using "__connection_header" field in ROS messages (for tf2_ros).


tfoote:
Merged into indigo-devel

PR #56, opened 2014-03-10
make can_transform code return bool

v4hn:
The current interface returns (bool, errorstr). This is nice to have, but it's
unintuitive and provokes the mistake "not buffer.can_transform(..)" which
should really be "not buffer.can_transform(..)[0]". Such a bug ( as currently
in tf2_ros/buffer.py:72 ) is much worse than receiving an error string without 
raising an exception.


tfoote:
Can't change a released packages API.

PR #57, opened 2014-03-10
make can_transform code return bool

v4hn:
The current interface returns (bool, errorstr). This is nice to have, but it's
unintuitive and provokes the mistake "not buffer.can_transform(..)" which
should really be "not buffer.can_transform(..)[0]". Such a bug ( as currently
in tf2_ros/buffer.py:72 ) is much worse than receiving an error string without 
raising an exception.

### DIFF
--- a/tf2_py/src/tf2_py.cpp
+++ b/tf2_py/src/tf2_py.cpp
@@ -201,10 +201,7 @@ static PyObject *canTransformCore(PyObject *self, PyObject *args, PyObject *kw)
 
   if (!PyArg_ParseTupleAndKeywords(args, kw, "ssO&", (char**)keywords, &target_frame, &source_frame, rostime_converter, &time))
     return NULL;
-  std::string error_msg;
-  bool can_transform = bc->canTransform(target_frame, source_frame, time, &error_msg);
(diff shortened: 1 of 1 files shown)


tfoote:
I agree that it's not optimal. However removing the ability to get the error message makes debugging impossible. It definitely should not be done at this level and propagate through all the changed APIs. 

There could be some options to add syntactic sugar or alternate APIs for the python interface.

v4hn:
I understand your reasoning, but I do not think debugging is that much of a problem here.
For most debugging purposes, one could afterwards simply use lookup_transform(timeout= Duration(0)) and look at
the resulting exception. The only "proper" alternatives I see here, are to add a class
CanTransformResult which defines **bool** accordingly or to add another function which
returns the tuple. Intuitively it just doesn't make sense to return a tuple in
a function named can_transform, especially because _tf's_ canTransform code doesn't do that.

What other ideas do you have in mind to change this in the long run?

tfoote:
To make this change we should change the Buffer class in buffer.py instead of the c bindings. The c bindings are not part of the public API.

tfoote:
closing in favor of #70

PR #58, opened 2014-03-10
Fix the can_transform design flaw at the other end.

v4hn:
As it's important to keep the API stable, let's fix
the only other end (mis)using the can_transform API
in the standard ROS distribution.

Also fix another bug which was never discovered due
to the broken can_transform check


tfoote:
Thanks

PR #62, opened 2014-05-12
Fix format string

trainman419:
It was interpreting the %20e as a format specifier and printing like:

```
[ WARN] [1399920996.445817921, 28.800000000]: TF_OLD_DATA ignoring data from the past for frame smooth at time 28.8 according to authority /simulator
Possible reasons are listed at http://wiki.ros.org/tf/Errors        8.000000e-01xplained
```


tfoote:
Thanks good catch.

Issue #64, opened 2014-05-16
Provide a way to clear tf2 buffers when the clock is reset


trainman419:
When running a simulation or playing back a bag file with --clock, I often try to leave most of my nodes running, and only restart the thing that is publishing to /clock

Unfortunately, the tf2 internal buffer does not reset when /clock jumps backwards, and the tf2 subscribers don't have a manual mechanism to tell them that the clock has been reset, and they should clear their buffers.

See http://answers.ros.org/question/164026/how-can-i-reset-the-simulated-clock/ for more information.

jbohren:
+100

v4hn:
+1 In theory there should be a mechanism in place to _automatically_ clean the buffers when a loop is detected. I know this worked with tf by EOL in fuerte because I fixed it myself.
Also, I remember, @tfoote afterwards pushed some related changes for tf2.
I didn't investigate, but I currently have problems with tf2 and rosbag play --loop as well (current hydro).

ZdenekM:
+1

skohlbr:
+1

tfoote:
duplicate #43

PR #68, opened 2014-06-27
Make transform_listener reset tf buffer on loop detection

v4hn:
This makes tf2 work with `rosbag --clock --loop` as expected.

For some reason rostime does not use CLOCK_MONOTONIC,
so I'm not sure whether we should allow for tiny backjumps (< 1s).
I know for sure this is an issue with the rospy API at least.

Also addresses issue #64.


v4hn:
Also, please cherry-pick this change to indigo-devel if this request is merged.

PR #70, opened 2014-09-18
python patch for #57

tfoote:
instead of patching lower level API expose the tuple to the user only if requested via optional argument

@v4hn please take a look at this alternative implementation


v4hn:
+1 Looks good to me.

This is a minor API break,
but if anyone else used that interface before,
they likely _mis_used it before anyway..

PR #71, opened 2014-09-28
add support for transforming sensor_msgs::PointCloud2

vrabaud:
How would something like that look ?


tfoote:
Thanks for this contribution! Sorry i didn't get to it earler.

vrabaud:
I wanted some feedback before merging it but sure :)
@mikeferguson , any feedback ?

tfoote:
As an independant package implementing the standard APIs it's pretty uncontroversial. I don't have a problem with the name etc. I"ve run into an issue with the unit tests I'm debugging. After some more testing I'll push out a release. There's still time for feedback.

PR #73, opened 2014-12-06
display difference between reference clock and each tf node, which was supported until groovy

k-okada:
frame_view until groovy shows reference clock and how much each node in tf tree differ from that[1], which is very convenient when we use ros over multiple machines and find out which machine needs to be ntpdate-ed. but from hydro, that information is not displayed[2].

This patch shows that information, but not work for simulation environment which we need to read '/clock' and if I understand correctly the tf2 is designed not to depend on ROS communication framework. But still this helps when we working on real robot.

[1]
screenshot from 2014-12-06 13 40 10

[2]
screenshot from 2014-12-06 13 42 16

### DIFF
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -170,6 +170,7 @@ BufferCore::BufferCore(ros::Duration cache_time)
   frameIDs_["NO_PARENT"] = 0;
   frames_.push_back(TimeCacheInterfacePtr());
   frameIDs_reverse.push_back("NO_PARENT");
+  ros::Time::init();
 }
 
 BufferCore::~BufferCore()
@@ -1023,6 +1024,7 @@ std::string BufferCore::allFramesAsYAML() const
   std::stringstream mstream;
   boost::mutex::scoped_lock lock(frame_mutex_);
 
+  ros::Time current_time = ros::Time::now();
   TransformStorage temp;
 
   if (frames_.size() ==1)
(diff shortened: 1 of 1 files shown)


tf2/src/buffer_core.cpp
```diff
@@ -170,6 +170,7 @@ BufferCore::BufferCore(ros::Duration cache_time)
   frameIDs_["NO_PARENT"] = 0;
   frames_.push_back(TimeCacheInterfacePtr());
   frameIDs_reverse.push_back("NO_PARENT");
+  ros::Time::init();
```
tfoote, line 173: This does not look safe. A BufferCore object could be constructed at any time during execution. Why are you adding this here?

k-okada, line 173: I need to call either ros::Time::init or ros::init to know current time using ros::Time::now() here -> https://github.com/ros/geometry_experimental/pull/73/files#diff-23f1780342d49d752e5fabba0fa7de1cR1329

tfoote, line 173: This module is specifically designed to be ROS independant as much as possible. Use of ros::Time was specifically designed out because it does not work in the case of a python wrapper. (ros::Time only works if roscpp is running. in pytf implementations ros::Time::now() does not work in simulated time) We need to find a way to implement this without adding that dependency back.

k-okada, line 173: Ok, could you give me any hint to get current time in tf2 library?

## 

◉ Kei Okada

tfoote, line 173: I think the best way would be to provide a second version of the function where you pass in the current time. If the new API is called it renders it with the time delta. If you use the old API it renders without the extra information. And then the higher level process(aka view_frames) will pass in ros::Time::now() to the BufferCore instance.

k-okada:
create new PR #74, I'm not sure if we should create new function `_allFramesAsDot2` or change `_allFramesAsDot` with optional argument.

```
-  std::string _allFramesAsDot() const;
+  std::string _allFramesAsDot(double current_time = 0) const;
```

tfoote:
Replaced by https://github.com/ros/geometry/issues/77 with #74

PR #74, opened 2014-12-18
change _allFramesAsDot() -> _allFramesAsDot(double current_time)

k-okada:


tfoote:
Thanks. I think the optional argument will be fine. I'll test this and merge before my next release.

PR #76, opened 2015-01-21
ODR violation fixes and more conversions 

vrabaud:


tfoote:
Thanks Vincent that looks good. I'll process it shortly.

vrabaud:
actually, I also need to add a transform for the QuaternionStamped, please hold on.

vrabaud:
I did add the types I needed but I'd like another one: being able to convert from geometry_msgs::Pose to tf2::Transform. It is actually done a lot in navigation but semantically, those two are different. Mathematically though, they are (rotation + translation). Can I add it ? (it's useful, otherwise you have to rewrite that conversion method a lot).

tfoote:
Yeah, I'd prefer the distinction but we need to use the same underlying datatype so adding the conversion seems fine. You're right people will do it/want to do it anyway.

vrabaud:
Hi, this PR is now good to go. Navigation has its tests passing. Thx.

PR #77, opened 2015-01-22
Fix static broadcaster with rpy args

paulbovbel:
static_transform_publisher was not setting the header properly for the RPY version of the input args.

Just like in tf(1), the parameters are provided backwards as x y z yaw pitch roll. Maybe a note should be added to specify that these are NOT Euler angles?

### DIFF
--- a/tf2_ros/src/static_transform_broadcaster_program.cpp
+++ b/tf2_ros/src/static_transform_broadcaster_program.cpp
@@ -81,7 +81,7 @@ int main(int argc, char ** argv)
     msg.transform.translation.z = atof(argv[3]);
 
     tf2::Quaternion quat;
-    quat.setRPY(atof(argv[3]), atof(argv[2]), atof(argv[1]));
+    quat.setRPY(atof(argv[6]), atof(argv[5]), atof(argv[4]));
     msg.transform.rotation.x = quat.x();
     msg.transform.rotation.y = quat.y();
     msg.transform.rotation.z = quat.z();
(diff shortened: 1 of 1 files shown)


tf2_ros/src/static_transform_broadcaster_program.cpp
```diff
@@ -78,22 +79,17 @@ int main(int argc, char ** argv)
...
-    double cosRoll = cos(halfRoll);
-    double sinRoll = sin(halfRoll);
-    
-    msg.transform.rotation.x = sinRoll * cosPitch * cosYaw - cosRoll * sinPitch * sinYaw;
-    msg.transform.rotation.y = cosRoll * sinPitch * cosYaw + sinRoll * cosPitch * sinYaw;
-    msg.transform.rotation.z = cosRoll * cosPitch * sinYaw - sinRoll * sinPitch * cosYaw;
-    msg.transform.rotation.w = cosRoll * cosPitch * cosYaw + sinRoll * sinPitch * sinYaw;
 
+    tf2::Quaternion quat;
+    quat.setRPY(atof(argv[3]), atof(argv[2]), atof(argv[1]));
```
tfoote: These should be 6,5,4.

tfoote:
Thanks, using the built in is much cleaner. One comment in line that it's using the wrong set of args.

tf2_ros/src/static_transform_broadcaster_program.cpp
paulbovbel: Thanks for the catch!

paulbovbel:
Fixed, thanks

paulbovbel:
Can this be merged in? tf2 static broadcaster not usable in the current state, frame_id's weren't being set

PR #78, opened 2015-01-22
Add sensor_msgs::Imu transform to tf2

paulbovbel:
Transforms IMU data created in an IMUs reference frame to the equivalent data in another frame. Particularly useful for getting data in an ENU frame from an NED IMU.

Notes:
- ignores pose, only keeps orientation of transform
- Transforming the orientation transforms the orientation's implicit reference frame as well (e.g. moving data from NED to ENU). Therefore the quaternion is projected into the new frame with new axes, rather than applying an offset like a typical transform.

### DIFF
--- a/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.h
+++ b/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.h
@@ -33,6 +33,7 @@
 #include <tf2/convert.h>
 #include <sensor_msgs/PointCloud2.h>
 #include <sensor_msgs/Imu.h>
+#include <sensor_msgs/MagneticField.h>
 #include <sensor_msgs/point_cloud2_iterator.h>
 #include <Eigen/Eigen>
 #include <Eigen/Geometry>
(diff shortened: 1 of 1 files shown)


paulbovbel:
Please hold pull, debugging

paulbovbel:
Tested and functional

paulbovbel:
Tested experimentally, would appreciate a math review.

mintar:
Looks like a really useful feature to have! Didn't test/review it properly though. Just an idea, if you want to cross-test this, Chad Rockey has written a node ([imu_transformer](https://github.com/ros-perception/imu_pipeline/blob/849e023530759c3d6bae62d132afecfc3cc85d1c/src/imu_transformer.cpp)) that does essentially the same thing, so perhaps that's useful to cross-validate your implementation. Still, I think this would be a great thing to have in tf2!

paulbovbel:
Added a sensor_msgs::MagneticField transform as well

tfoote:
Thanks these look useful. I'll try to review them soon.

tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.h
```diff
@@ -86,6 +88,151 @@ void fromMsg(const sensor_msgs::PointCloud2 &msg, sensor_msgs::PointCloud2 &out)
...
+      imu_in.linear_acceleration.x, imu_in.linear_acceleration.y, imu_in.linear_acceleration.z);
+
+
+  imu_out.linear_acceleration.x = accel.x();
+  imu_out.linear_acceleration.y = accel.y();
+  imu_out.linear_acceleration.z = accel.z();
+
+  transformCovariance(imu_in.linear_acceleration_covariance, imu_out.linear_acceleration_covariance, r);
+
+  Eigen::Quaternion<double> orientation = r * Eigen::Quaternion<double>(
```
hordurk, line 156: Shouldn't orientation be `imu_in.orientation*r.inverse()`, instead of `r*imu_in.orienation*r.inverse()` ?
I tried this code with an IMU mounted tilted on an object and `imu_in.orientation*r.inverse()` works for me (unless I'm misunderstanding what it should do. I had assumed when having an IMU mounted on an object with transform only being a rotation around y-axis, when the object it's mounted on is level pointing east, the output of transform should be unit quaternion?)

paulbovbel, line 156: You're definitely right. I thought that this code could do double duty  as an NED->ENU conversion, but I have messed up the offset-application for regular transforms, will need to rethink this, feel free to pull your change to my fork.

asimay:
sorry. how to use this?

mintar:
@asimay Please stop spamming unrelated issues with your question. You've already opened ccny-ros-pkg/imu_tools#63 , I'll answer there shortly.

nicolaje:
Hello,

what is the status of this? It is a very useful feature.

peci1:
@paulbovbel Would you remember why has this PR been closed without any comment? It still seems quite important to me (although there is workaround via the imu_transformer package).

PR #81, opened 2015-02-03
Remove unused assignments in max/min functions

jonbinney:
While running clang's static analysis on my own code, it complains (correctly) about unused assignments in tf2. These assignments to maxVal/minVal are unnecessary because they never get read after this point.


tfoote:
Thanks @jonbinney

Issue #80, opened 2015-02-02
message_filter behavior with "/" frames


vrabaud:
Just wondering whether it's a design decision or a bug.

With tf2, your frames cannot start with "/". Good.

Now, I have an old ROS bag (from navigation) where frames do start with "/". There is a tf2_ros::message_filter on those but it assumes the frames do not start with "/". This results in miscellaneous errors.

What is the correct behavior ?
- this bag is wrong and cannot work with tf2 so the current behavior is right
- tf2_ros::message_filter should clean the frames before using tf2. I actually fixed it and that works but I am not sure you want to allow such a loose behavior.

Let me know if that is more appropriate and I'll make a PR

PR #83, opened 2015-03-08
filters: fix unsupported old messages with frame_id starting with '/'

vrabaud:
This is useful for navigation that has tests relying on old rosbags
whose messages have frames starting with '/' but I believe it is
useful to anybody for backward compatibility

Connects to #80

### DIFF
--- a/tf2_ros/include/tf2_ros/message_filter.h
+++ b/tf2_ros/include/tf2_ros/message_filter.h
@@ -242,7 +242,8 @@ public:
   {
     boost::mutex::scoped_lock frames_lock(target_frames_mutex_);
 
-    target_frames_ = target_frames;
+    target_frames_.resize(target_frames.size());
+    std::transform(target_frames.begin(), target_frames.end(), target_frames_.begin(), this->stripSlash);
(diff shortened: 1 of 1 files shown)


vrabaud:
any problem with that one ?

tfoote:
tf2 is supposed to not use '/' prefixed names. We tried to design tf2 to not be doing this sort of backwards compatibility logic. The glue layers from tf to tf2 all strip the slashes conditionally, so the logic is only in the backwards compatability layers. New code in tf2 should never encounter frame_ids with slashes. Recorded bagfiles are a problem, as the frame_ids are baked in. It makes sense for backwards compatibility to strip the slash on receipt. Are the target frames sent from bag files?

vrabaud:
yop, they are. Then again, I see some slash stripping in tf2: https://github.com/ros/geometry_experimental/blob/indigo-devel/tf2/src/buffer_core.cpp#L107
It's just that message_filters are not as core as buffer_core and I believe they got forgotten.
(just rebased btw)

tfoote:
Yeah, lets make it happen. The stripping can always be removed at some much later date. Thanks for the rebase.

PR #86, opened 2015-03-12
Use overloading instead of default arguments

### LINKED ISSUE #84: undefined reference to `tf2::BufferCore::_allFramesAsDot() const'
after #74 I get this error
/opt/ros/indigo/lib/libtf.so: undefined reference to `tf2::BufferCore::_allFramesAsDot() const'

to reproduce: create a workspace with the ros/velodyne repo and the ros/geometry_experimental repo, both of them checked out to the latest version.

I bisected on ros/geometry_experimental and I confirmed #74 is the culprit

tfoote:
This has better backwards compatibility for symbols. 
Fixes #84

### DIFF
--- a/tf2/include/tf2/buffer_core.h
+++ b/tf2/include/tf2/buffer_core.h
@@ -212,7 +212,11 @@ public:
   /** \brief A way to see what frames have been cached in yaml format
    * Useful for debugging tools
    */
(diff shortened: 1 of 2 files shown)


tfoote:
Tested by @bricerebsamen in #84

Issue #52, opened 2014-02-03
tf2_ros: Python documentation missing


jack-oquin:
It looks like there was an attempt to move the Python documentation from the tf2 package to tf2_ros, but that activity was not completed.

The fix is probably not difficult. I notice that `index.rst` is missing and also the rosdoc element in package.xml: 

``` xml
  <export>
    <rosdoc config="rosdoc.yaml" />
  </export>
```

The Sphinx `conf.py` looks wrong, too. If I can find some time, I'll submit a pull request.

jack-oquin:
I still don't see the Python docs on the wiki. Is it waiting for the build farm to update?

tfoote:
the online docs take up to 3 days to refresh. I'll try to accelerate this one

PR #96, opened 2015-04-03
Enabled build of tf2 python documentation

jackieokay:
Enable Sphinx documentation for tf2_ros for issue #52 and removed dead `opencv.willowgarage.com` link from `tf2_ros/conf.py` to speed up documentation generation. `rosdoc_lite tf2_ros` now generates a `python/tf2_ros.html` file from `index.rst` and `tf2_ros.rst`. The documentation could still probably use some work, and I'm not sure whether or not the directory structure needs to be changed on the docs server. @tfoote can advise.

Connects to #52


tfoote:
Looks good to me. It's already configured for python generation on the server so that should be fine.

PR #97, opened 2015-04-09
create tf2_eigen.

terakoji:
http://wiki.ros.org/tf2/Migration mention tf2_eigen.
But there are not implementations.

I really want to use tf2 with eigen.

Notes:
- only geometry_msgs::Pose and geometry_msgs::Point.


tfoote:
Thanks, this is a good start for eigen support. If you or anyone else wants to add mappings that would be great.

Issue #91, opened 2015-04-02
deadlock


bricerebsamen:
I migrated some of my code to tf2 and I'm getting some deadlocks. I am running the latest version from source, which included @vrabaud 's fix #79 

Here is the backtrace

```
#0  __lll_lock_wait ()
    at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007ffff540b657 in _L_lock_909 ()
   from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007ffff540b480 in __GI___pthread_mutex_lock (mutex=0x7ffffffefba8)
    at ../nptl/pthread_mutex_lock.c:79
#3  0x0000000000431368 in pthread_mutex_lock (m=0x7ffffffefba8)
    at /usr/include/boost/thread/pthread/mutex.hpp:61
#4  boost::mutex::lock (this=0x7ffffffefba8)
    at /usr/include/boost/thread/pthread/mutex.hpp:113
#5  0x00000000004313db in boost::unique_lock<boost::mutex>::lock (
    this=0x7ffffffeed90) at /usr/include/boost/thread/lock_types.hpp:346
#6  0x00007ffff7211e11 in boost::unique_lock<boost::mutex>::unique_lock(boost::mutex&) () from /home/brice/code/workspaces/merged/devel/lib/libtf2_ros.so
#7  0x00007ffff6ebea9f in tf2::BufferCore::cancelTransformableRequest(unsigned long) () from /home/brice/code/workspaces/merged/devel/lib/libtf2.so
#8  0x0000000000441e0c in tf2_ros::MessageFilter<pcl::PointCloud<stdr_velodyne::PointType> >::add (this=0x7fffffff00c0, evt=...)
    at /home/brice/code/workspaces/merged/src/driving/external/geometry_experimental/tf2_ros/include/tf2_ros/message_filter.h:381
#9  0x0000000000437eed in operator() (a0=..., this=0x6d2d28)
    at /usr/include/boost/function/function_template.hpp:767
#10 message_filters::CallbackHelper1T<ros::MessageEvent<pcl::PointCloud<stdr_velodyne::PointType> const> const&, pcl::PointCloud<stdr_velodyne::PointType> >::call (this=0x6d2d20, event=..., nonconst_force_copy=<optimized out>)
    at /opt/ros/indigo/include/message_filters/signal1.h:76
[...]

PR #101, opened 2015-04-17
switch to use a shared lock with upgrade instead of only a unique lock for message_mutex_

tfoote:
Connects to #91


bricerebsamen:
I can confirm that it fixes the deadlock reported in #91

tfoote:
Great thanks for the verification!

PR #104, opened 2015-04-26
fix bad math for Vector3

vrabaud:
my bad ........


vrabaud:
actually, wasn't my fault: it was there from the very beginning, as shown by the tests and #106 :)

PR #106, opened 2015-05-09
fix test for Vector3

vrabaud:
The test is failing because of the recent fix in the header.
The values have to be the same as Point anyway.

### DIFF
--- a/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
+++ b/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
@@ -74,7 +74,7 @@ template <>
 inline
   void doTransform(const geometry_msgs::Vector3Stamped& t_in, geometry_msgs::Vector3Stamped& t_out, const geometry_msgs::TransformStamped& transform)
   {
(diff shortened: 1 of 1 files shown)


vrabaud:
it seems that this test and the math that went with it only applied the rotation of a transformation (and not its translation). Was I wrong fixing the code in #104 and this test ?

tfoote:
Yes, upon review the lines changed in #104 should be reverted. It should only apply the rotation to a "Vector" for the full transform use the Point methods. https://github.com/ros/geometry_experimental/pull/104/files#diff-bb6bd2d938a7c0ed4397920b3d645fd3L116

vrabaud:
Agreed. fixing with #109 
It all happened because I was porting some code to tf2 and it should have used points in the first place ... Sorry for the noise.

vrabaud:
confusion came from the difference of convention of a Vector3 in tf2 (vs geometry_msgs): https://github.com/ros/geometry_experimental/blob/indigo-devel/tf2/include/tf2/LinearMath/Transform.h#L91
In tf2, it's not just a direction (which geometry_msgs does not state clearly btw ....)

tfoote:
In most core math libraries there's not a distinction, the tf2 LinearMath is just a fork of Bullet's LinearMath. We should update the docs for the Vector vs Point message to be clearer.

PR #111, opened 2015-05-17
provide more conversions between types

vrabaud:
The previous conversion always assumed that it was converting a
non-message type to a non-message type. Now, one, both or none
can be a message or a non-message.

### DIFF
--- a/test_tf2/test/test_convert.cpp
+++ b/test_tf2/test/test_convert.cpp
@@ -38,6 +38,7 @@
 #include <tf2/convert.h>
 #include <tf2_kdl/tf2_kdl.h>
 #include <tf2_bullet/tf2_bullet.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <ros/time.h>
 
 TEST(tf2Convert, kdlToBullet)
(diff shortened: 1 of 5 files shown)


vrabaud:
@Karsten1987, that's what you need right ?

tfoote:
@vrabaud, this gets rid of the need to define toMsg and fromMsg like here? https://github.com/ros/geometry_experimental/blob/indigo-devel/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h#L85

vrabaud:
yes. Now, that would break the API (maybe somebody is using those functions somewhere ...). Do you want me to remove them from tf2_*_msgs ?
I was thinking about creating templated toMsg / fromMsg, specialized for messages using SFINAE, to remove those functions for messages. That would not break the API. But the first one is easier so please let's break it :)
In case you want me to go with the second one, can I use C++11 ? std::enable_if would help there.

vrabaud:
ok, to properly remove the toMsg/fromMsg functions you define, I created a templated version of them with a specialized version for messages. I still wonder why they were not templates in the first place ... (how are you supposed to know that you need to implement them ?)

vrabaud:
I fixed your 4 comments and updated that PR

vrabaud:
and updated again sorry :)

vrabaud:
BTW, I am planning to specialize getTimestamp and getFrame for messages too in another pull request, any objection ? It just seems that more messages are being added to tf2 these days, so it's just less work to do.

tfoote:
Great, the template is a lot cleaner. We mostly focused on getting the existing datatypes working and weren't focused on a generic converter at the time. 

generic getTimestamp and getFrame seems good to me

vrabaud:
cleaned up some comments, anything else ?

vrabaud:
any doubts on this one ? I'd like to move on with the getYaw and such.

tfoote:
Thanks this looks good. Sorry for the delay reviewing.

PR #112, opened 2015-05-17
Get yaw

vrabaud:
This needs #111 to be in but it provides a few useful functions when using tf2. Maybe we should have a new header like utils.h or something.

### DIFF
--- a/test_tf2/CMakeLists.txt
+++ b/test_tf2/CMakeLists.txt
@@ -27,6 +27,9 @@ target_link_libraries(test_tf2_message_filter ${Boost_LIBRARIES} ${catkin_LIBRAR
 catkin_add_gtest(test_convert test/test_convert.cpp)
 target_link_libraries(test_convert ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${orocos_kdl_LIBRARIES})
 
+catkin_add_gtest(test_utils test/test_utils.cpp)
+target_link_libraries(test_utils ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${orocos_kdl_LIBRARIES})
+
 add_executable(test_buffer_server EXCLUDE_FROM_ALL test/test_buffer_server.cpp)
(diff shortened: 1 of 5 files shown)


vrabaud:
@Karsten1987, you must be so happy to have such a boss :)

vrabaud:
TODO: get it to work with Stamped version of the data. Will probably wait for #111 to be merged before updating.

tfoote:
Thanks @vrabaud that looks like a good approach. Maybe euler.h would be a good file name? Keep it relatively clearly scoped.

vrabaud:
well, there is also getIdentity in there.

vrabaud:
how about a utils.h that would contain getTimestamp, getFrameId, getYaw, getEulerYPR, getIdentity and have convert.h include it (for backward compatiblity) and only contain doTransform, toMsg, fromMsg and convert ?

vrabaud:
ok, updated. @tfoote , @mikeferguson I need some advice before validating that:
- I restricted getEulerYPR and get Yaw to work with anything that can be converted to a quaternion. hence, it would not work for a Pose and such. Should I specialize for things that contain a rotation ? (I personnally think it would be wrong and it's not such a hassle to write `getEulerYPR(t.quaternion)`
- getIdentity only works for anything that can be converted to a `tf2::Transform`. Should I name it differently ? Should I specialize it for anything that has some identity (like Vector3, Quaternion) ?

mikeferguson:
@vrabaud on the first point, that is exactly what I would expect, getEulerYPR(t.quaternion) is very clear what the intent is, and is not much more to write.

On the getIdentity() -- I don't really have a strong opinion either way. I think it is fine as-is, and if it becomes a problem later, we can add additional specializations.

tfoote:
I agree with @mikeferguson . 

This looks pretty good. Anything else you're thinking @vrabaud ?

vrabaud:
all I did was rename `getIdentity` to `getTransformIdentity` as I believe we might need `getQuaternionIdentity` and a few other ones.

So, all good to me.

vrabaud:
that will be released after the indigo and jade syncs I guess ?

PR #113, opened 2015-06-01
Revert "officialize toMsg and fromMsg by adding them to convert"

vrabaud:
This reverts commit ea87ea91e67b5f1ed28134cafeea8a1c6329e7ec.

We just don't want to break the ABI

### DIFF
--- a/test_tf2/test/test_convert.cpp
+++ b/test_tf2/test/test_convert.cpp
@@ -95,25 +95,7 @@ TEST(tf2Convert, kdlBulletROSConversions)
   EXPECT_NEAR(b1.x(), b4.x(), epsilon);
   EXPECT_NEAR(b1.y(), b4.y(), epsilon);
   EXPECT_NEAR(b1.z(), b4.z(), epsilon);
-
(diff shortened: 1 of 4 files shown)


vrabaud:
Hi, sorry for the inconvenience but with the latest ABI breakage in image_pipeline, I'd rather not get that commit in, cf http://www.ros.org/reps/rep-0009.html

It is just a cleanup anyway: for K-turtle, we might as well remove the toMsg and fromMsg for message functions (as they are already handled by the `convert` function).

mikeferguson:
tf2 is not >= version 1.0 though, right? so ABI breakage is fine, right? We break the ABI/API of moveit all the time....

tfoote:
Since we're not making guarantees about ABI we don't need to be super strict. 

Can we do it where we add the automatic methods which is just adding symbols but only remove the old style methods later?

vrabaud:
ok, I only reverted the toMsg and fromMsg for messages as they don't make sense.

tfoote:
looks good. thanks. I'll create a jade branch to push this back into.

PR #114, opened 2015-06-24
Add KDL::Frame to TransformStamped conversion

paulbovbel:
Follows the convention of the existing inverse TransformStamp to KDL::Frame conversion function

### DIFF
--- a/tf2_kdl/include/tf2_kdl/tf2_kdl.h
+++ b/tf2_kdl/include/tf2_kdl/tf2_kdl.h
@@ -50,6 +50,16 @@ KDL::Frame transformToKDL(const geometry_msgs::TransformStamped& t)
 		      KDL::Vector(t.transform.translation.x, t.transform.translation.y, t.transform.translation.z));
   }
 
+inline
+geometry_msgs::TransformStamped kdlToTransform(const KDL::Frame& k)
+{
+  geometry_msgs::TransformStamped t;
+  t.transform.translation.x = k.p.x();
+  t.transform.translation.y = k.p.y();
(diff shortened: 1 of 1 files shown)


tfoote:
The analog between a `kdl::Frame` is a `geometry_msgs::Transform` not a `TransformStamped`. The method you've duplicated was intended to be a local helper function not a public API.

A better solution would be to extend the automatic conversions here to support `geometry_msgs::TransformStamped` to `tf2::Stamped<kdl::Frame>`. It should be basically a copy and paste of this section: https://github.com/ros/geometry_experimental/blob/jade-devel/tf2_kdl/include/tf2_kdl/tf2_kdl.h#L172-L207

From the dependent diff, you're using a TransformStamped anyway. 

If you want to implement the lower level conversion and reuse it, I think it would be better to use place the method here: https://github.com/ros/geometry_experimental/blob/jade-devel/tf2_kdl/include/tf2_kdl/tf2_kdl.h#L46 in parallel with the `transformToKDL` method.

paulbovbel:
Sadly this:
`geometry_msgs::PoseStamped toMsg(const tf2::Stamped<KDL::Frame>& in)`

 is already implemented, so I cannot reuse the signature to convert to TransformStamped.

I have moved the helper method to tf2_kdl.h, is this acceptable?

paulbovbel:
ping @tfoote, could you please merge and release when you have a chance to look things over?

tfoote:
This looks good. I'll merge and work toward a coordinated release with nodelets and possibly ros_comm.

PR #126, opened 2015-11-16
add missing Python runtime dependency

vrabaud:
https://github.com/ros/geometry_experimental/pull/119 from @laurent-george revealed a bug that was there before (a missing runtime dependency):
http://jenkins.ros.org/job/devel-indigo-geometry_experimental/51/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/testReport/junit/nose.failure/Failure/runTest/


tfoote:
thanks for the attention @vrabaud

PR #128, opened 2015-12-09
Vector3 not affected by translation in a transform

### LINKED ISSUE #124: transforming a Vector3 in python applies the translation
according to the doc:

```
# This represents a vector in free space. 
# It is only meant to represent a direction. Therefore, it does not
# make sense to apply a translation to it (e.g., when applying a 
# generic rigid transformation to a Vector3, tf2 will only apply the
# rotation). If you want your data to be translatable too, use the
# geometry_msgs/Point message instead.
```

(from http://docs.ros.org/jade/api/geometry_msgs/html/msg/Vector3.html)

the following code shows that it's not the case:

```
t = TransformStamped()
t.transform.translation.z = 1
t.transform.rotation.w = 1

v = Vector3Stamped()
v.vector.x = 1
v.vector.y = 0
v.vector.z = 0

vt = tf2_geometry_msgs.do_transform_vector3(v, t)
```

As a result, `vt.z` is 1, so the translation is applied.

chapulina:
Fixes #124 
Connects to #130


chapulina:
I made the test script into a unit test like in pull request #132

tfoote:
lgtm

PR #133, opened 2015-12-15
casted el to string in view_frames

g-gemignani:


tfoote:
This looks reasonable. Can you share a use case where this is a problem?

g-gemignani:
Hello,
This is a problem when you have a frame that uses a number as name.
For example, in our project we had a frame called "54". This made view_frames.py crash. 
I understand that this is a corner case but I believe that it might still happen.

g-gemignani:
Hello,
as suggested I created a pull request for throwing an error when a not unitary quaternion is given in input to the static_transform_publisher

tfoote:
I've cherry picked the first commit.

The check for normalization needs to have some tolerance. Expecting the value to be normalized to full double floating point precision is going to fail many things which are effectively normalized. It will take some research to determine what an appropriate tolerance would be.

tfoote:
normalization followed up in #196

PR #135, opened 2015-12-16
Create a version of tf2 for ros2.

tfoote:
It is tested up to static_transform_publisher and tf2_echo and tf2_monitor are working.

### DIFF
--- a/test_tf2/CHANGELOG.rst
+++ b/test_tf2/CHANGELOG.rst
@@ -97,7 +97,7 @@ Changelog for package test_tf2
 ------------------
 * splitting rospy dependency into tf2_py so tf2 is pure c++ library.
 * switching to console_bridge from rosconsole
-* moving convert methods back into tf2 because it does not have any ros dependencies beyond ros::Time which is already a dependency of tf2
+* moving convert methods back into tf2 because it does not have any ros dependencies beyond builtin_interfaces::msg::Time which is already a dependency of tf2
 * Cleaning up unnecessary dependency on roscpp
 * converting contents of tf2_ros to be properly namespaced in the tf2_ros namespace
 * Cleaning up packaging of tf2 including:
--- a/test_tf2/CHANGELOG.rst
+++ b/test_tf2/CHANGELOG.rst
@@ -97,7 +97,7 @@ Changelog for package test_tf2
 ------------------
 * splitting rospy dependency into tf2_py so tf2 is pure c++ library.
 * switching to console_bridge from rosconsole
-* moving convert methods back into tf2 because it does not have any ros dependencies beyond builtin_interfaces::msg::Time which is already a dependency of tf2
+* moving convert methods back into tf2 because it does not have any ros dependencies beyond ros::Time which is already a dependency of tf2
 * Cleaning up unnecessary dependency on roscpp
 * converting contents of tf2_ros to be properly namespaced in the tf2_ros namespace
 * Cleaning up packaging of tf2 including:
--- a/test_tf2/test/buffer_core_test.cpp
+++ b/test_tf2/test/buffer_core_test.cpp
@@ -38,7 +38,7 @@
 
 void seed_rand()
 {
-  //Seed random number generator with current microseond count
+  //Seed random number generator with current time.
   srand(std::chrono::system_clock::now().time_since_epoch().count());
 };
 
--- a/test_tf2/test/buffer_core_test.cpp
+++ b/test_tf2/test/buffer_core_test.cpp
@@ -162,7 +162,7 @@ void push_back_1(std::vector<std::string>& children, std::vector<std::string>& p
   dy.push_back(0.0);
 }
 
-void setupTree(tf2::BufferCore& mBC, const std::string& mode, const builtin_interfaces::msg::Time & time, const tf2::TempDuration& interpolation_space = tf2::TempDuration())
+void setupTree(tf2::BufferCore& mBC, const std::string& mode, const builtin_interfaces::msg::Time & time, const tf2::Duration& interpolation_space = tf2::Duration())
 {
   ROS_DEBUG("Clearing Buffer Core for new test setup");
   mBC.clear();
@@ -237,7 +237,7 @@ void setupTree(tf2::BufferCore& mBC, const std::string& mode, const builtin_inte
         else
           ts.child_frame_id = frames[i]; // connect first frame
         EXPECT_TRUE(mBC.setTransform(ts, "authority"));
-        if (interpolation_space > tf2::TempDuration())
+        if (interpolation_space > tf2::Duration())
         {
           ts.header.stamp = time + interpolation_space * .5;
           EXPECT_TRUE(mBC.setTransform(ts, "authority"));
@@ -276,7 +276,7 @@ void setupTree(tf2::BufferCore& mBC, const std::string& mode, const builtin_inte
     ts.header.frame_id = parents[i];
     ts.child_frame_id = children[i];
     EXPECT_TRUE(mBC.setTransform(ts, "authority"));
-    if (interpolation_space > tf2::TempDuration())
+    if (interpolation_space > tf2::Duration())
     {
       ts.header.stamp = time + interpolation_space * .5;
       EXPECT_TRUE(mBC.setTransform(ts, "authority"));
@@ -632,11 +632,11 @@ TEST(BufferCore_lookupTransform, i_configuration)
   builtin_interfaces::msg::Time eval_time;
   permuter.addOptionSet(times, &eval_time);
 
-  std::vector<tf2::TempDuration> durations;
-  durations.push_back(tf2::TempDuration(1.0));
-  durations.push_back(tf2::TempDuration(0.001));
-  durations.push_back(tf2::TempDuration(0.1));
-  tf2::TempDuration interpolation_space;
+  std::vector<tf2::Duration> durations;
+  durations.push_back(tf2::Duration(1.0));
+  durations.push_back(tf2::Duration(0.001));
+  durations.push_back(tf2::Duration(0.1));
+  tf2::Duration interpolation_space;
   //  permuter.addOptionSet(durations, &interpolation_space);
 
   std::vector<std::string> frames;
@@ -970,11 +970,11 @@ TEST(BufferCore_lookupTransform, one_link_configuration)
   builtin_interfaces::msg::Time eval_time;
   permuter.addOptionSet(times, &eval_time);
 
-  std::vector<tf2::TempDuration> durations;
-  durations.push_back(tf2::TempDuration(1.0));
-  durations.push_back(tf2::TempDuration(0.001));
-  durations.push_back(tf2::TempDuration(0.1));
-  tf2::TempDuration interpolation_space;
+  std::vector<tf2::Duration> durations;
+  durations.push_back(tf2::Duration(1.0));
+  durations.push_back(tf2::Duration(0.001));
+  durations.push_back(tf2::Duration(0.1));
+  tf2::Duration interpolation_space;
   //  permuter.addOptionSet(durations, &interpolation_space);
 
   std::vector<std::string> frames;
@@ -1014,11 +1014,11 @@ TEST(BufferCore_lookupTransform, v_configuration)
   builtin_interfaces::msg::Time eval_time;
   permuter.addOptionSet(times, &eval_time);
 
-  std::vector<tf2::TempDuration> durations;
-  durations.push_back(tf2::TempDuration(1.0));
-  durations.push_back(tf2::TempDuration(0.001));
-  durations.push_back(tf2::TempDuration(0.1));
-  tf2::TempDuration interpolation_space;
+  std::vector<tf2::Duration> durations;
+  durations.push_back(tf2::Duration(1.0));
+  durations.push_back(tf2::Duration(0.001));
+  durations.push_back(tf2::Duration(0.1));
+  tf2::Duration interpolation_space;
   //  permuter.addOptionSet(durations, &interpolation_space);
 
   std::vector<std::string> frames;
@@ -1061,11 +1061,11 @@ TEST(BufferCore_lookupTransform, y_configuration)
   builtin_interfaces::msg::Time eval_time;
   permuter.addOptionSet(times, &eval_time);
 
-  std::vector<tf2::TempDuration> durations;
-  durations.push_back(tf2::TempDuration(1.0));
-  durations.push_back(tf2::TempDuration(0.001));
-  durations.push_back(tf2::TempDuration(0.1));
-  tf2::TempDuration interpolation_space;
+  std::vector<tf2::Duration> durations;
+  durations.push_back(tf2::Duration(1.0));
+  durations.push_back(tf2::Duration(0.001));
+  durations.push_back(tf2::Duration(0.1));
+  tf2::Duration interpolation_space;
   //  permuter.addOptionSet(durations, &interpolation_space);
 
   std::vector<std::string> frames;
@@ -1107,11 +1107,11 @@ TEST(BufferCore_lookupTransform, multi_configuration)
   builtin_interfaces::msg::Time eval_time;
   permuter.addOptionSet(times, &eval_time);
 
-  std::vector<tf2::TempDuration> durations;
-  durations.push_back(tf2::TempDuration(1.0));
-  durations.push_back(tf2::TempDuration(0.001));
-  durations.push_back(tf2::TempDuration(0.1));
-  tf2::TempDuration interpolation_space;
+  std::vector<tf2::Duration> durations;
+  durations.push_back(tf2::Duration(1.0));
+  durations.push_back(tf2::Duration(0.001));
+  durations.push_back(tf2::Duration(0.1));
+  tf2::Duration interpolation_space;
   //  permuter.addOptionSet(durations, &interpolation_space);
 
   std::vector<std::string> frames;
@@ -1319,10 +1319,10 @@ TEST(BufferCore_lookupTransform, helix_configuration)
 
     tf2::BufferCore mBC;
 
-    builtin_interfaces::msg::Time     t0        = builtin_interfaces::msg::Time() + tf2::TempDuration(10);
-    tf2::TempDuration step      = tf2::TempDuration(0.05);
-    tf2::TempDuration half_step = tf2::TempDuration(0.025);
-    builtin_interfaces::msg::Time     t1        = t0 + tf2::TempDuration(5.0);
+    builtin_interfaces::msg::Time     t0        = builtin_interfaces::msg::Time() + tf2::Duration(10);
+    tf2::Duration step      = tf2::Duration(0.05);
+    tf2::Duration half_step = tf2::Duration(0.025);
+    builtin_interfaces::msg::Time     t1        = t0 + tf2::Duration(5.0);
 
     /*
      * a->b->c
@@ -1435,11 +1435,11 @@ TEST(BufferCore_lookupTransform, ring_45_configuration)
   builtin_interfaces::msg::Time eval_time;
   permuter.addOptionSet(times, &eval_time);
 
-  std::vector<tf2::TempDuration> durations;
-  durations.push_back(tf2::TempDuration(1.0));
-  durations.push_back(tf2::TempDuration(0.001));
-  durations.push_back(tf2::TempDuration(0.1));
-  tf2::TempDuration interpolation_space;
+  std::vector<tf2::Duration> durations;
+  durations.push_back(tf2::Duration(1.0));
+  durations.push_back(tf2::Duration(0.001));
+  durations.push_back(tf2::Duration(0.1));
+  tf2::Duration interpolation_space;
   //  permuter.addOptionSet(durations, &interpolation_space);
 
   std::vector<std::string> frames;
@@ -2020,7 +2020,7 @@ TEST(tf, Exceptions)
 
 TEST(tf, NoExtrapolationExceptionFromParent)
 {
-  tf::Transformer mTR(true, tf2::TempDuration().fromNSec(1000000));
+  tf::Transformer mTR(true, tf2::Duration().fromNSec(1000000));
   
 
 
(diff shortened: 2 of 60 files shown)


tf2/CMakeLists.txt
```diff
@@ -1,54 +1,53 @@
...
 )
 
 # Tests
-if(CATKIN_ENABLE_TESTING)
+if(AMENT_ENABLE_TESTING)
+find_package(ament_lint_auto REQUIRED)
+ament_lint_auto_find_test_dependencies()
+   
+ament_add_gtest(test_cache_unittest test/cache_unittest.cpp)
+target_link_libraries(test_cache_unittest tf2 ${geometry_msgs_LIBRARIES} ${console_bridge_LIBRARIES})
```
dirk-thomas: This must check if the gtest target actually exists.

Same below.

tf2/CMakeLists.txt
```diff
@@ -1,54 +1,53 @@
...
-target_link_libraries(test_simple tf2  ${console_bridge_LIBRARIES})
-add_dependencies(test_simple tf2_msgs_gencpp)
-
-add_executable(speed_test EXCLUDE_FROM_ALL test/speed_test.cpp)
-target_link_libraries(speed_test tf2  ${console_bridge_LIBRARIES})
-add_dependencies(tests speed_test)
-add_dependencies(tests tf2_msgs_gencpp)
+# add_executable(speed_test EXCLUDE_FROM_ALL test/speed_test.cpp)
+# target_link_libraries(speed_test tf2  ${geometry_msgs_LIBRARIES} ${console_bridge_LIBRARIES})
+# add_dependencies(tests speed_test)
```
dirk-thomas, line 55: Why are those tests disabled? Are they supposed to be reenabled in the future? Should we add a TODO for it?

test_tf2/CHANGELOG.rst
```diff
@@ -97,7 +97,7 @@ Changelog for package test_tf2
 ------------------
 * splitting rospy dependency into tf2_py so tf2 is pure c++ library.
 * switching to console_bridge from rosconsole
-* moving convert methods back into tf2 because it does not have any ros dependencies beyond ros::Time which is already a dependency of tf2
+* moving convert methods back into tf2 because it does not have any ros dependencies beyond builtin_interfaces::msg::Time which is already a dependency of tf2
```
wjwwood: I don't think you should change the existing changelog entries.

test_tf2/test/buffer_core_test.cpp
```diff
@@ -39,9 +39,7 @@
 void seed_rand()
 {
   //Seed random number generator with current microseond count
-  timeval temp_time_struct;
-  gettimeofday(&temp_time_struct,NULL);
-  srand(temp_time_struct.tv_usec);
+  srand(std::chrono::system_clock::now().time_since_epoch().count());
```
wjwwood, line 42: Does the ratio (microseconds vs nanoseconds) matter here?

test_tf2/test/buffer_core_test.cpp
```diff
@@ -164,7 +162,7 @@ void push_back_1(std::vector<std::string>& children, std::vector<std::string>& p
   dy.push_back(0.0);
 }
 
-void setupTree(tf2::BufferCore& mBC, const std::string& mode, const ros::Time & time, const ros::Duration& interpolation_space = ros::Duration())
+void setupTree(tf2::BufferCore& mBC, const std::string& mode, const builtin_interfaces::msg::Time & time, const tf2::TempDuration& interpolation_space = tf2::TempDuration())
```
wjwwood: If you have a placeholder type for Duration, shouldn't there be one for Time as well? For example `builtin_interfaces::msg::Time` doesn't respect sim time, and if this code is meant to work with sim time at all, it will need to be replaced in the future.

tf2/package.xml
```diff
@@ -16,19 +17,17 @@
...
-  <run_depend>libconsole-bridge-dev</run_depend>
-  <run_depend>geometry_msgs</run_depend>
-  <run_depend>rostime</run_depend>
-  <run_depend>tf2_msgs</run_depend>
+  <depend>console_bridge</depend>
+  <depend>geometry_msgs</depend>
+  
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <!-- <test_depend>ament_lint_common</test_depend> -->
```
dirk-thomas: Without any linters `ament_lint_auto` and the corresponding CMake code doesn't make much sense.

tf2_msgs/CMakeLists.txt
```diff
@@ -1,20 +1,28 @@
...
+rosidl_generate_interfaces(${PROJECT_NAME}
+  "msg/TF2Error.msg"
+  "msg/TFMessage.msg"
+  "srv/FrameGraph.srv"
+  DEPENDENCIES builtin_interfaces std_msgs geometry_msgs
+)
+#  actionlib_msgs
 
+# TODO(tfoote) Need amend_export_dependenceies?
+ament_export_dependencies(rosidl_generator_cpp)
```
dirk-thomas: You won't need this if you only want to export the generated messages of this package.

tf2_msgs/package.xml
```diff
@@ -10,17 +11,18 @@
...
 
-  <run_depend>actionlib_msgs</run_depend>
-  <run_depend>geometry_msgs</run_depend>
-  <run_depend>message_generation</run_depend>
+  <!-- <depend>actionlib_msgs</depend> -->
+  <depend>builtin_interfaces</depend>
+  <depend>geometry_msgs</depend>
 
-</package>
+  <exec_depend>rosidl_default_runtime</exec_depend>
```
dirk-thomas, line 22: I don't think such a package exists currently.

tf2_ros/CMakeLists.txt
```diff
@@ -1,97 +1,156 @@
...
+#   actionlib
+#   actionlib_msgs
+#   geometry_msgs
+#   message_filters
+#   roscpp
+#   rosgraph
+#   tf2
+#   tf2_msgs
+#   ${TF2_PY}
+# )
```
dirk-thomas: Should these catkin-specific comments be removed?

tf2_ros/CMakeLists.txt
```diff
@@ -1,97 +1,156 @@
...
-  rosgraph
-  tf2
-  tf2_msgs
-  ${TF2_PY}
-)
-
-include_directories(include ${catkin_INCLUDE_DIRS})
+find_package(ament_cmake REQUIRED)
+# find_package(actionlib REQUIRED)
+# find_package(actionlib_msgs REQUIRED)
```
dirk-thomas, line 15: Adding a TODO to reenable disabled parts.

tf2/CHANGELOG.rst
```diff
@@ -134,7 +134,7 @@ Changelog for package tf2
 ------------------
 * splitting rospy dependency into tf2_py so tf2 is pure c++ library.
 * switching to console_bridge from rosconsole
-* moving convert methods back into tf2 because it does not have any ros dependencies beyond ros::Time which is already a dependency of tf2
+* moving convert methods back into tf2 because it does not have any ros dependencies beyond builtin_interfaces::msg::Time which is already a dependency of tf2
```
wjwwood: Same changelog comment.

tf2_ros/CMakeLists.txt
```diff
@@ -1,97 +1,156 @@
...
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  DESTINATION include/${PROJECT_NAME}
 )
 
 
 # Tests
-if(CATKIN_ENABLE_TESTING)
-
+# if(AMENT_ENABLE_TESTING)
+if (false)
```
dirk-thomas, line 123: Why has this part not been updated? This needs a TODO at least.

tf2_msgs/package.xml
tfoote, line 22: There is not such package. I just was exploring that for ros2/rclcpp#181. 

I copied this from common_interfaces and rcl_interfaces. Should we be removing it everywhere? @wjwwood mentioned it might be necessary for packaging but if we don't have a need for it now we should probably drop it everywhere.

tf2/CMakeLists.txt
```diff
@@ -1,54 +1,53 @@
...
-add_dependencies(tf2 tf2_msgs_gencpp)
+target_link_libraries(tf2 ${geometry_msgs_LIBRARIES} ${console_bridge_LIBRARIES})
 
 install(TARGETS tf2
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION lib/${PROJECT_NAME}
```
wjwwood: Why not `bin` here? I think it might be important for it to be `bin` here because Windows puts `.dll`'s in the `RUNTIME DESTINATION`.

tf2_ros/include/tf2_ros/buffer_interface.h
```diff
@@ -103,75 +124,77 @@ class BufferInterface
...
+  // template <class A, class B>
+  //   B& transform(const A& in, B& out, 
+	// 	 const std::string& target_frame, const tf2::TimePoint& target_time,
+	// 	 const std::string& fixed_frame, tf2::TempDuration timeout=tf2::TempDuration(0.0)) const
+  // {
+  //   // do the transform
+  //   A copy = transform(in, target_frame, target_time, fixed_frame, timeout);
+  //   tf2::convert(copy, out);
+  //   return out;
+  // }
```
dirk-thomas: Please use block comments to avoid such unnecessary diffs.

tf2/include/tf2/buffer_core.h
```diff
@@ -109,7 +110,20 @@ class BufferCore
...
+                                                 transform.transform.rotation.y,
+                                                 transform.transform.rotation.z),
+                                 tf2::Vector3(transform.transform.translation.x,
+                                              transform.transform.translation.y,
+                                              transform.transform.translation.z));
+    TimePoint time_point(std::chrono::nanoseconds(transform.header.stamp.nanosec) +
+        std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::seconds(transform.header.stamp.sec)));
+    return setTransformImpl(tf2_transform, transform.header.frame_id, transform.child_frame_id,
+                            time_point, authority, is_static);
+  }
```
wjwwood: Why was the implementation moved here? Does it need to be `inline` since it's not a template?

wjwwood: Same for several below.

dirk-thomas:
The patch shows how invasive the types `Time` and `Duration` are being used. Instead of simply replacing the ROS 1 types with ROS 2 equivalents I would suggest to create at least typedefs for them in a single place and even better implement functions to operate on them. Then the whole code base only uses the custom defined types and the functions and allows to easily change this.

tf2/include/tf2/convert.h
```diff
@@ -32,10 +32,14 @@
 #ifndef TF2_CONVERT_H
 #define TF2_CONVERT_H
 
+#include <builtin_interfaces/msg/time.hpp>
+
+// TODO(tfoote) 
+#error This has not been converted to ROS2 and will not compile
```
wjwwood, line 38: Is this needed for ros2 support or is this ok to overlook in this pr?

tf2/include/tf2/exceptions.h
```diff
@@ -33,9 +33,21 @@
 #define TF2_EXCEPTIONS_H
 
 #include <stdexcept>
+#include <cstdint>
 
 namespace tf2{
 
+enum class TF2Error : std::uint8_t{
```
wjwwood: nitpick: `enum class TF2Error : std::uint8_t {` use a whitespace before the `{`

tf2/package.xml
wjwwood: Should this be removed?

tf2/src/buffer_core.cpp
```diff
@@ -41,60 +40,7 @@
...
-{
-  tx.translation.x = 0;
-  tx.translation.y = 0;
-  tx.translation.z = 0;
-  tx.rotation.x = 0;
-  tx.rotation.y = 0;
-  tx.rotation.z = 0;
-  tx.rotation.w = 1;
-}
+// constexpr TempDuration BufferCore::DEFAULT_CACHE_TIME;
```
wjwwood: Should be removed?

tf2/test/cache_unittest.cpp
```diff
@@ -45,7 +42,7 @@ void seed_rand()
   values.clear();
   for (unsigned int i = 0; i < 1000; i++)
   {
-    int pseudo_rand = std::floor(i * M_PI);
+    int pseudo_rand = std::floor(i * 3.141592653589793);
```
wjwwood, line 45: That sucks, but I can't find a better way to reference it portably.

tf2_ros/CMakeLists.txt
```diff
@@ -1,97 +1,156 @@
...
+find_package(ament_cmake REQUIRED)
+# find_package(actionlib REQUIRED)
+# find_package(actionlib_msgs REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_msgs REQUIRED)
+
+find_package(rmw)
+find_package(rmw_implementation) # not needed after rmw typesupport refactor
```
wjwwood: Still needed?

tf2_ros/CMakeLists.txt
```diff
@@ -1,97 +1,156 @@
...
+
+include_directories(include
+  ${geometry_msgs_INCLUDE_DIRS}
+  ${rclcpp_INCLUDE_DIRS}
+  ${rmw_INCLUDE_DIRS}
+  ${rmw_implementation_INCLUDE_DIRS}
+  ${tf2_INCLUDE_DIRS}
+  ${tf2_msgs_INCLUDE_DIRS}
+  )
+message(fatal_error "tf2_msgs_INCLUDE_DIRS ${tf2_msgs_INCLUDE_DIRS}")
```
wjwwood: ?

tf2_ros/CMakeLists.txt
```diff
@@ -1,97 +1,156 @@
...
+  DESTINATION include/${PROJECT_NAME}
 )
 
 
 # Tests
-if(CATKIN_ENABLE_TESTING)
-
+# if(AMENT_ENABLE_TESTING)
+if (false)
+
```
wjwwood: All tests are disabled?

wjwwood: Should add a TODO to know when they should be added back, i.e. what is preventing them from working now.

tf2_ros/include/tf2_ros/buffer.h
```diff
@@ -112,21 +112,22 @@ namespace tf2_ros
...
   private:
-    bool getFrames(tf2_msgs::FrameGraph::Request& req, tf2_msgs::FrameGraph::Response& res) ;
+    bool getFrames(tf2_msgs::srv::FrameGraph::Request& req, tf2_msgs::srv::FrameGraph::Response& res) ;
 
 
     // conditionally error if dedicated_thread unset.
     bool checkAndErrorDedicatedThreadPresent(std::string* errstr) const;
 
-    ros::ServiceServer frames_server_;
+//TODO(renable framegraph service)
```
wjwwood: nitpik: `// TODO(tfoote) re-enable frame graph service`, also what is a frame graph service (for my own edification)?

tfoote: The frame graph service allows you to query a specific instance of the listener for it's graph remotely instead of listening at your instance. It's very useful for debugging remote instances, so you can see the state  (or usually lack there of) built up inside any listener.

tf2_ros/include/tf2_ros/transform_listener.h
```diff
@@ -32,14 +32,13 @@
...
 #define TF2_ROS_TRANSFORMLISTENER_H
 
-#include "std_msgs/Empty.h"
-#include "tf2_msgs/TFMessage.h"
-#include "ros/ros.h"
-#include "ros/callback_queue.h"
+#include <thread>
+#include "tf2_msgs/msg/tf_message.hpp"
+#include "rclcpp/rclcpp.hpp"
+// #include "ros/callback_queue.h"
```
wjwwood: delete this?

tf2_ros/package.xml
```diff
@@ -11,38 +11,31 @@
...
+  <depend>tf2_msgs</depend>
+
+  <!-- <run_depend>message_filters</run_depend>
   <run_depend>rosgraph</run_depend>
   <run_depend>rospy</run_depend>
   <run_depend>std_msgs</run_depend>
   <run_depend>tf2</run_depend>
   <run_depend>tf2_msgs</run_depend>
   <run_depend>tf2_py</run_depend>
+  <test_depend>rostest</test_depend> -->
```
wjwwood: Should these ROS 1 dependencies be left in as comments? I'd say drop them.

tf2_ros/src/buffer_server_main.cpp
```diff
@@ -61,7 +61,7 @@ int main(int argc, char** argv)
   nh.param("buffer_size", buffer_size, 120.0);
 
   // WIM: this works fine:
-  tf2_ros::Buffer buffer_core(ros::Duration(buffer_size+0)); // WTF??
+  tf2_ros::Buffer buffer_core(tf2::TempDuration(buffer_size+0)); // WTF??
```
wjwwood: lol

wjwwood:
Well, other than my comments I guess I'd +1 this for merge.

However, it still needs a lot of work and it is a bit discouraging that we don't have tests to verify that things behave as expected. There are a bunch of workarounds in this pr which just change the way `tf` works and I guess if the goal is to get something merged then that's ok, but realistically without some underlying features hashed out in ROS 2 this is pretty hack-y, e.g. latching, logging, ROS Time (sim time), etc.

tf2/CMakeLists.txt
tfoote, line 55: todo added with comment as to why it's removed.

test_tf2/CHANGELOG.rst
tfoote: To aggressive global search and replace. Reverted here and in a few other places.

test_tf2/test/buffer_core_test.cpp
tfoote, line 42: No, it's just needs a random number. Comment updated.

tf2/CMakeLists.txt
tfoote: This is according to the migration guide: http://design.ros2.org/articles/migration_guide_from_ros1.html 

@dirk-thomas thoughts?

tf2/include/tf2/convert.h
tfoote, line 38: This is currently commented out of the build as the datatype conversions packages were not included. This is how you interact with them. You have to use either the message datatypes or the tf2 native datatypes.

tf2/CMakeLists.txt
dirk-thomas: I agree with @wjwwood . `CATKIN_PACKAGE_BIN_DESTINATION` was wrong in the first place and wouldn't have worked on Windows.

tf2_ros/CMakeLists.txt
tfoote: left over debugging, gone

tf2_ros/package.xml
tfoote: I removed all the ported ones, and todo's the remaining ones as to be added back when ported.

tf2_ros/CMakeLists.txt
tfoote: TODO added with comment.

tf2/package.xml
tfoote: I commented both and added a TODO to enable linters.

tf2_msgs/CMakeLists.txt
tfoote: Removed

tf2_msgs/package.xml
tfoote, line 22: Conveniently we just added this package in https://github.com/ros2/rosidl/pull/57

tf2_ros/CMakeLists.txt
tfoote: Stripped

tf2/CMakeLists.txt
tfoote: I've updated it in both packages and added it to my update of the migration guide.

tf2_ros/CMakeLists.txt
tfoote, line 15: added todos

tf2_ros/CMakeLists.txt
tfoote, line 123: TODO added. They are close to being able to be ported. They use some things that are not available yet such as sim_time.

tf2_ros/include/tf2_ros/buffer_interface.h
tfoote: Done

tf2/src/buffer_core.cpp
tfoote: Removed

tf2/include/tf2/buffer_core.h
tfoote: It was supposed to be a very thin wrapper, but the datatypes are complex enough to warrant moving. Both moved.

tf2/CMakeLists.txt
tfoote: Added

tfoote:
I created the aliases for tf2::TimePoint and tf2::Duration and deployed them through the compiling code. I also cleaned up a few windowsisms necessary to build. 

CI:
http://ci.ros2.org/job/ci_linux/803/
http://ci.ros2.org/job/ci_osx/671/ (expecting test failures due to ros2/ros2#174)
http://ci.ros2.org/job/ci_windows/857 (The bullet test fails)

PR #137, opened 2015-12-19
remove the use of slashes from topics

tfoote:
DDS does not support slashes. 

We have held back on developing our integrated naming design with the potential to alias ROS topics and DDS topic names.

### DIFF
--- a/tf2_ros/src/static_transform_broadcaster.cpp
+++ b/tf2_ros/src/static_transform_broadcaster.cpp
@@ -43,7 +43,7 @@ StaticTransformBroadcaster::StaticTransformBroadcaster()
   rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
   custom_qos_profile.depth = 100;
   // TODO(tfoote) latched equivalent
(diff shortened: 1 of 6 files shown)


gerkey:
+1

mikeferguson:
Out of curiosity -- when you say "does not support slashes" -- is that mean just at the beginning of the name, or throughout? That is, will we be unable to "namespace" things, such as is currently done with many drivers, like "camera/rgb/image" and "camera/rgb/image_raw" ??

wjwwood:
This is just a temporary limitation. We have to decide on the policy for addressing in ROS 2 before committing to any solution in the code.

tfoote:
http://ci.ros2.org/job/ci_linux/818

wjwwood:
DDS does not allow `/` in the topics at all, but we can use another separator in place of `/`. We intend to support `/` in the ROS API's as well as namespacing, just not yet.

mikeferguson:
thanks @wjwwood -- I wasn't sure what the story was and that was a bit concerning

PR #139, opened 2015-12-26
Remove LGPL from license tags

jspricke:
LGPL was erroneously included in 2a38724. As there are no files with it
in the package, let's remove the tag as well.


tfoote:
Thanks for the cleanup.

PR #140, opened 2016-01-04
fixed linker error in echo and monitor due to missing pthread

firesurfer:
Added -pthread in CMakeLists.txt in tf2_ros so it compiles on a current debian testing with gcc (Debian 5.3.1-4) 5.3.1 20151219

### DIFF
--- a/tf2_ros/CMakeLists.txt
+++ b/tf2_ros/CMakeLists.txt
@@ -6,7 +6,7 @@ if(NOT WIN32)
 endif()
 
 if(NOT ANDROID)
-set(TF2_PY tf2_py)
+  set(TF2_PY tf2_py)
 endif()
 
 find_package(ament_cmake REQUIRED)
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks for the report and proposed fix. Two comments. One is that there's a deeper issue in that tf_echo does not use threads directly so it should be getting that flag exported from a lower level package. And unfortunately `-pthread` is not windows compatible so we can't merge this as is.

firesurfer:
I changed the CMakeLists.txt so it will link to pthread only on non windows systems.

gerkey:
It does look like c8c9a19 will make the change portable, but I agree with @tfoote that we should fix the underlying issue (in some lower package, either export `-pthread` or remove the use of pthread).

@firesurfer, are you able to determine from the failed link where the use of pthread is introduced?

jackieokay:
Victor also reported a link error in this package and in intra_process_demo on Debian Jessie with armhf: https://github.com/ros2/common_interfaces/issues/11#issuecomment-186451997

I haven't done any investigation into this but I suspect this is related to usage of `std::thread` and//or `std::condition_variable` in rclcpp...

firesurfer:
Oh yeah, I forgot about the error in the demos because I always remove them from the build. 
It's definitely the use of std::thread which uses internally -pthread. Some systems (Ubuntu apparently too) automatically link against pthread and some other libs. Debian doesn't.  

Unfortunately I don't know the tf2 code good enough to find the part the uses pthread without doing a lot of digging in the code.

PR #142, opened 2016-01-11
Added eigenTotransform function

boris-il-forte:
There was no eigenToTransform function as the equivalent kdlToTransform tf2_kdl.

The function does not fill the transform information (child frame and header information), as the kdl equivalent does.

should be interesting to add to the signature of the function the missing information needed to get a complete TransformStamped message

### DIFF
--- a/tf2_eigen/include/tf2_eigen/tf2_eigen.h
+++ b/tf2_eigen/include/tf2_eigen/tf2_eigen.h
@@ -44,6 +44,23 @@ Eigen::Affine3d transformToEigen(const geometry_msgs::TransformStamped& t) {
 					      t.transform.rotation.x, t.transform.rotation.y, t.transform.rotation.z));
 }
 
+inline
+geometry_msgs::TransformStamped eigenToTransform(const Eigen::Affine3d& T)
+{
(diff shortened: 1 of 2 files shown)


boris-il-forte:
I strongly suggest to merge this pull request as it does not only add a missing interface function, but also fix a compilation issue if header is included in multiple files (missing inline keyword)

boris-il-forte:
any news about this issue? the missing `inline` keyword seemes to affect not only the jade branch, but also the indigo branch. This makes tf_eigen header not usable for including in multiple files.

Also the pull request adds an useful function, that is implemented in tf_kdl

tfoote:
This looks reasonable. I've got several other projects with higher priority. If you could add unit tests for the new function that would make reviewing it much easier.

boris-il-forte:
Added test case. I hope is a sufficient one...

boris-il-forte:
Any news about this?
If you need better unit test just tell me...

tfoote:
Thanks for this as well as adding the unit test. And the inline methods will be valuable.

tfoote:
Closing in favor of #158

PR #150, opened 2016-01-27
Add python_orocos_kdl dependency for ros-*-ros-base installs

daenny:
We were missing PyKDL when using tf2_geometry_msgs, when ros is only installed in its ros-*-ros-base version and not desktop full version.
This should fix it.


tfoote:
Thanks. Yes tf2_geometry_msgs.py needs these dependencies.

PR #151, opened 2016-02-06
Only set c++11 flag for C++

esteve:
Otherwise this will fail with C code (such as the Python extension) on OSX:

http://ci.ros2.org/job/ci_osx/766/console


tfoote:
lgtm

PR #152, opened 2016-02-09
fix documentation warnings

vrabaud:
Those would happen when doing rosdoc_lite ./


tfoote:
Thanks for the cleanup!

tfoote:
cherry-picked to indigo 72fa60575aad885bd1de289240f7b4b20d0a418b

PR #153, opened 2016-02-10
cleanup for windows warnings

tfoote:
This cleans up most warnings for tf2. 

The ctime fix is not quite working but I want to get this in for now and can return to that later. I think it needs the `#define` and explicit include here: http://en.cppreference.com/w/c/chrono/ctime

CI: http://ci.ros2.org/job/ci_windows/1021/

### DIFF
--- a/tf2/include/tf2/time.h
+++ b/tf2/include/tf2/time.h
@@ -56,14 +56,12 @@ namespace tf2
     return durationToSec(Duration(timepoint.time_since_epoch()));
   }
 
-  // Display functions as there is no default display
-  // TODO: find a proper way to handle display
   inline std::string displayTimePoint(const TimePoint& stamp)
   {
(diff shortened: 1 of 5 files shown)


tfoote:
@esteve can you review?

esteve:
I attempted to fix the ctime issue in https://github.com/ros/geometry_experimental/pull/136 using the same code, but unfortunately it doesn't seem to solve anything :-(

http://ci.ros2.org/job/ci_windows/1021/warnings34Result/package.1841019875/

tfoote:
Yeah, I'd suggest we merge this as it stands. It fixes all but the two ctime windows warnings from the one line and come back to that later.

esteve:
@tfoote +1, merge it and let's iterate to fix the remaining warnings.

esteve:
I'll close the other pull request.

PR #136, opened 2015-12-18
Fix Windows warnings

esteve:
This fixes several Windows warnings

PR #154, opened 2016-02-13
Get rid of date/time formatting, just use double.

tfoote:
This will make better sense when using sim time anyway.

Fixes #136 

CI: 
http://ci.ros2.org/job/ci_linux/997
http://ci.ros2.org/job/ci_osx/828
http://ci.ros2.org/job/ci_windows/1077

### DIFF
--- a/tf2/include/tf2/time.h
+++ b/tf2/include/tf2/time.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2016, Open Source Robotics Foundation, Inc.
+ * Copyright (c) 2015-2016, Open Source Robotics Foundation, Inc.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -31,8 +31,10 @@
  #define TF2_TIME_H
 
 #include <chrono>
+#include <new>
 #include <stdio.h>
 #include <string>
+#include <string.h>
 #include <thread>
 
 namespace tf2
@@ -64,12 +66,18 @@ namespace tf2
     const char * format_str = "%.6f";
     double current_time = timeToSec(stamp);
     int buff_size = snprintf(NULL, 0, format_str, current_time);
+    if (buff_size < 0) {
+      throw std::runtime_error(strerror(errno));
+    }
     char * buffer = new char[buff_size];
-    snprintf(buffer, buff_size, format_str, current_time);
+    int bytes_written = snprintf(buffer, buff_size, format_str, current_time);
+    if (bytes_written < 0) {
+      throw std::runtime_error(strerror(errno));
+    }
     std::string result = std::string(buffer);
     delete[] buffer;
     return result;
   }
 }
 
-#endif // TF2_TIME_CACHE_H
+#endif // TF2_TIME_H
--- a/tf2/include/tf2/time.h
+++ b/tf2/include/tf2/time.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, Open Source Robotics Foundation, Inc.
+ * Copyright (c) 2015, 2016, Open Source Robotics Foundation, Inc.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -27,9 +27,12 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+ #ifndef TF2_TIME_H
+ #define TF2_TIME_H
+
 #include <chrono>
-#include <iomanip>
-#include <ctime>
+#include <stdio.h>
+#include <string>
 #include <thread>
 
 namespace tf2
@@ -50,7 +53,7 @@ namespace tf2
   inline double durationToSec(const tf2::Duration & input){
     return (double)std::chrono::duration_cast<std::chrono::seconds>(input).count();
   }
-  
+
   inline double timeToSec(const TimePoint& timepoint)
   {
     return durationToSec(Duration(timepoint.time_since_epoch()));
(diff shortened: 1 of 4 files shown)


tfoote:
@esteve can you review?  I've just ended up switching to snprintf. The warnings are gone on windows: http://ci.ros2.org/job/ci_windows/1077/warnings34Result/

tf2/include/tf2/time.h
```diff
@@ -58,10 +58,9 @@ namespace tf2
 
   inline std::string displayTimePoint(const TimePoint& stamp)
   {
-    std::time_t time = std::chrono::system_clock::to_time_t(std::chrono::time_point_cast<std::chrono::milliseconds>(stamp));
-    char str[100];
-    std::strftime(str, sizeof(str), "%c", std::localtime(&time));
-    return std::string(str);
+    char buffer [100];
```
dirk-thomas: This buffer could be smaller.

tfoote: It could be smaller, apparently depending on local etc 25 for a double known not to be long enough. I wanted to guarantee that we'd have enough space. I erred on the side of caution since bytes on the stack for a quick function call are relatively cheap. What do you suggest for the size of the buffer?  

I could also make it dynamically allocate the buffer to keep it to exactly the right size. https://stackoverflow.com/a/3923207/604099 but the extra complexity and computational cost of that seems overkill.

dirk-thomas: I first thought it was using `%d` and then the buffer could be smaller. But its actually `%f` which mean 100 characters might not be enough depending on the value. I don't think that a hard coded value will do for `%f` a double. Also in case it does get truncated I think this function need to signal the error case.

tfoote:
Changed to make the buffer dynamic. 

CI triggered:
http://ci.ros2.org/job/ci_linux/999/
http://ci.ros2.org/job/ci_osx/830/
http://ci.ros2.org/job/ci_windows/1079/

tf2/include/tf2/time.h
```diff
@@ -58,11 +61,15 @@ namespace tf2
...
+    snprintf(buffer, buff_size, format_str, current_time);
+    std::string result = std::string(buffer);
+    delete[] buffer;
+    return result;
   }
-
-
 }
+
+#endif // TF2_TIME_CACHE_H
```
dirk-thomas: Doesn't match header guard above.

tf2/include/tf2/time.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, Open Source Robotics Foundation, Inc.
+ * Copyright (c) 2015, 2016, Open Source Robotics Foundation, Inc.
```
dirk-thomas: 2015-2016

tf2/include/tf2/time.h
```diff
@@ -58,11 +61,15 @@ namespace tf2
 
   inline std::string displayTimePoint(const TimePoint& stamp)
   {
-    std::time_t time = std::chrono::system_clock::to_time_t(std::chrono::time_point_cast<std::chrono::milliseconds>(stamp));
-    char str[100];
-    std::strftime(str, sizeof(str), "%c", std::localtime(&time));
-    return std::string(str);
+    const char * format_str = "%.6f";
+    double current_time = timeToSec(stamp);
+    int buff_size = snprintf(NULL, 0, format_str, current_time);
```
dirk-thomas: The return value needs to be checked for > 0 and error cases <= 0 need to be handled.

tf2/include/tf2/time.h
```diff
@@ -58,11 +61,15 @@ namespace tf2
...
+    char * buffer = new char[buff_size];
```
dirk-thomas: Need to check `buffer` for successful allocation.

tfoote:
Fixups from review: CI on linux and osx passed, windows waiting for executor. 

http://ci.ros2.org/job/ci_linux/1001/
http://ci.ros2.org/job/ci_osx/832/
http://ci.ros2.org/job/ci_windows/1082

tf2/include/tf2/time.h
dirk-thomas: This hasn't been addressed yet.

tf2/src/time.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+std::string tf2::displayTimePoint(const TimePoint& stamp)
+{
+  const char * format_str = "%.6f";
+  double current_time = timeToSec(stamp);
+  int buff_size = snprintf(NULL, 0, format_str, current_time);
+  if (buff_size < 0) {
+#ifdef _WIN32
+    size_t errmsglen = strerrorlen_s(errno) + 1;
+    char errmsg[errmsglen];
```
dirk-thomas: I don't expect Visual Studio to eat that: https://msdn.microsoft.com/en-us/library/zb1574zs(v=vs.140).aspx

tf2/include/tf2/time.h
tfoote: new throws bad_alloc when it fails: http://www.cplusplus.com/reference/new/bad_alloc/

tf2/src/time.cpp
tfoote: Fixed and canceled and retriggered: http://ci.ros2.org/job/ci_windows/1083/

tf2/include/tf2/time.h
dirk-thomas: Oh right, back in C++ land :wink:

tfoote:
The warnings are fixed now: http://ci.ros2.org/job/ci_windows/1084/warnings34Result/fixed/

PR #157, opened 2016-02-18
work around console bridges odd lib path

tfoote:
Test build: http://ci.ros2.org/view/packaging/job/packaging_linux/150

### DIFF
--- a/tf2_ros/CMakeLists.txt
+++ b/tf2_ros/CMakeLists.txt
@@ -20,6 +20,12 @@ find_package(tf2_msgs REQUIRED)
 
 find_package(rmw)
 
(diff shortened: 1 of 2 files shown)


tfoote:
Successfully tested against the binaries produced by 150 above.

wjwwood:
+1

Please create an issue on this repository which says to undo this once we figure out something more permanent with upstream `console_bridge`.

Also please add a comment about why we need to do this in the `CMakeLists.txt` so future me will have some context :smile:.

tfoote:
Created tracking ticket and added comments to the cmake and env hook file with a link to the issue.

PR #162, opened 2016-03-23
Fix comment on BufferCore::MAX_GRAPH_DEPTH.

gavanderhoorn:
Noticed it while following the link to `buffer_core.h` via [Is there a max number of edges allowed between two frames in a TF tree?](http://answers.ros.org/question/229902/is-there-a-max-number-of-edges-allowed-between-two-frames-in-a-tf-tree/) on ROS Answers.


gavanderhoorn:
Should I reopen against `jade-devel`?

tfoote:
No this is the right place. jade-devel was prematurely created we haven't forked development.

PR #167, opened 2016-04-02
Fix indexing beyond end of array

eric-wieser:


tfoote:
Thanks for catching that.

PR #170, opened 2016-04-09
tf2: Add tf2_geometry_msgs dependency

kartikmohta:
`tf2/include/tf2/impl/utils.h` includes `tf2_geometry_msgs/tf2_geometry_msgs.h` but the dependency is missing

### DIFF
--- a/tf2/package.xml
+++ b/tf2/package.xml
@@ -22,11 +22,13 @@
   <build_depend>libconsole-bridge-dev</build_depend>
   <build_depend>geometry_msgs</build_depend>
   <build_depend>rostime</build_depend>
+  <build_depend>tf2_geometry_msgs</build_depend>
   <build_depend>tf2_msgs</build_depend>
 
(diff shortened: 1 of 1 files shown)


tfoote:

tfoote:
Thanks for catching that. It is a problem. However adding this dependency creates a circular dependency since `tf2_geometry_msgs` depends on `tf2` thus we should move the implementation into `tf2_geometry_msgs` since it cannot be used without that package anyway.

kartikmohta:
Yeah agreed, seems like some restructuring is required so I'll let you guys do it the right way :)

tfoote (changes requested):
As mentioned in the comment this causes an issue with circular depedencies. I'm going to mark this as changes requested for tracking purposes.

tfoote:
I'm going to close this as there's now a ticket tracking it in #275 with a discussion of a better way to resolve this.

Issue #50, opened 2014-01-15
python3 compatability issue


tfoote:
From @focs at https://github.com/ros-gbp/geometry_experimental-release/issues/1

Hi,
I'm trying to compile ros on my gentoo system. I got stuck while trying to compile the tf2 package:

./src/catkin/bin/catkin_make_isolated --pkg tf2 --force-cmake

The error is:

[100%] Building CXX object CMakeFiles/tf2_py.dir/src/tf2_py.cpp.o
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:62:1: error: invalid conversion from 'const char_' to 'Py_ssize_t {aka long int}' [-fpermissive]
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp: In function 'PyObject_ transform_converter(const TransformStamped_)':
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:109:103: error: 'PyString_FromString' was not declared in this scope
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp: In function 'PyObject_ allFramesAsYAML(PyObject_, PyObject_)':
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:186:59: error: 'PyString_FromString' was not declared in this scope
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp: In function 'PyObject\* allFramesAsString(PyObject_, PyObject_)':
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:192:61: error: 'PyString_FromString' was not declared in this scope
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp: In function 'PyObject\* setTransform(PyObject_, PyObject_)':
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:388:105: error: 'PyString_AsString' was not declared in this scope
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp: At global scope:
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:457:1: warning: converting to non-pointer type 'int' from NULL [-Wconversion-null]
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp: In function 'void init_tf2()':
/home/fox/ros_catkin_ws/src/tf2/src/tf2_py.cpp:479:67: error: 'PyString_FromString' was not declared in this scope
[...]

PR #173, opened 2016-05-06
Make tf2_py compatible with python3.

de-vri-es:
This PR makes tf2_py compatible with python3. I compiled the package for both python 3.5 and 2.7 (which worked), but I **did not** test the resulting python module.

There were three issues to fix:
- Python3 uses unicode strings for text. The old python2 string type no longer exists.
- Module initialization has been overhauled for python3.
- Type definitions changed slightly.

I put wrapper functions for the correct string types in `python_compat.h` (and one convenience function for importing modules). Something similar could be done for the type definition and the module initialization, but those would involve writing new macros, which would not necessarily make the code easier to maintain (especially in the case of module initialization).

I did not use `py3c` (a C API compatibility layer for python2/3) to avoid adding a new dependency. If a new dependency is acceptable, `py3c` would be a good alternative. It would involve adding a new rosdep key though, I assume.

If this is the way to go, `PyObject_BorrowAttrString` could also be moved to `python_compat.h`, to group python convenience functions together. (Though I would rename it so it doesn't look like a real python function. It took me a while of digging through the python C API before realising it is defined in the source file itself).

### DIFF
--- a/tf2_py/src/python_compat.h
+++ b/tf2_py/src/python_compat.h
@@ -43,4 +43,11 @@ inline PyObject *pythonImport(const std::string & name)
   return module;
 }
 
+inline PyObject *pythonBorrowAttrString(PyObject* o, const char *name)
+{
+    PyObject *r = PyObject_GetAttrString(o, name);
+    Py_XDECREF(r);
+    return r;
+}
+
 #endif
--- a/tf2_py/src/python_compat.h
+++ b/tf2_py/src/python_compat.h
@@ -1,0 +1,46 @@
+#ifndef TF2_PY_PYTHON_COMPAT_H
+#define TF2_PY_PYTHON_COMPAT_H
+
+#include <Python.h>
+
+#include <string>
+
+inline PyObject *stringToPython(const std::string &input)
+{
+#if PY_MAJOR_VERSION >= 3
+  return PyUnicode_FromStringAndSize(input.c_str(), input.size());
+#else
+  return PyString_FromStringAndSize(input.c_str(), input.size());
+#endif
+}
+
+inline PyObject *stringToPython(const char *input)
+{
+#if PY_MAJOR_VERSION >= 3
+  return PyUnicode_FromString(input);
+#else
+  return PyString_FromString(input);
+#endif
+}
+
+inline std::string stringFromPython(PyObject * input)
+{
+  Py_ssize_t size;
+  char * data;
+#if PY_MAJOR_VERSION >= 3
+  data = PyUnicode_AsUTF8AndSize(input, &size);
+#else
+  PyString_AsStringAndSize(input, &data, &size);
+#endif
+  return std::string(data, size);
+}
+
+inline PyObject *pythonImport(const std::string & name)
+{
+  PyObject *py_name = stringToPython(name);
+  PyObject *module  = PyImport_Import(py_name);
+  Py_XDECREF(py_name);
+  return module;
+}
+
+#endif
--- a/tf2_py/src/tf2_py.cpp
+++ b/tf2_py/src/tf2_py.cpp
@@ -58,9 +58,13 @@ struct buffer_core_t {
 
 
 static PyTypeObject buffer_core_Type = {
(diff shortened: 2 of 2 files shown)


de-vri-es:
This should fix #50, btw.

de-vri-es:
One more addendum: Rewriting the whole thing in Cython seems like a good idea to me, perhabs even with a more full-featured API (while maintaining backwards compatibility of course).

However, I wanted python3 compatibility fast, and this was the quickest, least intrusive way to get that.

tfoote:

Thanks for the contribution. I'll try to review it tomorrow. Python3 support will be great.

tfoote:
Thanks for taking the time to do this. It looks good to me. Bumping `PyObject_BorrowAttrString` into the `python_compat.h` makes sense to me. 

Thanks for working around it without the extra dependency. If this is enough lets stick with this and not completely refactor it to Cython.

de-vri-es:
Moved `PyObject_BorrowAtrrString` to `python_compat.h` and renamed it to `pythonBorrowAttrString` to avoid looking like it is part of the official Python API.

(Update: And made it inline, since it is defined in a header.)

Thanks for the review.

tfoote:
Thanks again!

spaghetti-:
Thank you for this patch!

koniarik:
Hi folks, will there be anytime soon release of geometry2 with  this patch?

nstiurca:
@SquirrelCZE I'm wondering the same thing.

In the meantime, I built from source based on the latest version I just pulled (439e2350ca9b2f0581a3631f0929acc2021db73e), but I can't figure out how to actually use TF2 in Python 3. Here's what I did:

```
mkdir -p overlay_ws/src
cd overlay_ws/src
git clone https://github.com/ros/geometry2.git
catkin_init_workspace
cd ..
rosdep install --from-paths src --os=ubuntu:xenial
catkin_make --use-ninja --cmake-args -DCMAKE_BUILD_TYPE=Release
source devel/setup.bash
```

Then I tried to use it with

```
$ python3
Python 3.5.2 (default, Sep 10 2016, 08:21:44) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tf2_py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nicu/dbot/overlay_ws/devel/lib/python2.7/dist-packages/tf2_py/__init__.py", line 35, in <module>
    exec(__fh.read())
  File "<string>", line 37, in <module>
ImportError: No module named '_tf2'
>>> 
```

nstiurca:
I should say the import error is the same (except for paths) I get when only working with the latest version installed from APT.

MohamadHalwani:
> @SquirrelCZE I'm wondering the same thing.
> 
> In the meantime, I built from source based on the latest version I just pulled ([439e235](https://github.com/ros/geometry2/commit/439e2350ca9b2f0581a3631f0929acc2021db73e)), but I can't figure out how to actually use TF2 in Python 3. Here's what I did:
> 
> ```
> mkdir -p overlay_ws/src
> cd overlay_ws/src
> git clone https://github.com/ros/geometry2.git
> catkin_init_workspace
> cd ..
> rosdep install --from-paths src --os=ubuntu:xenial
> catkin_make --use-ninja --cmake-args -DCMAKE_BUILD_TYPE=Release
> source devel/setup.bash
> ```
> 
> Then I tried to use it with
> 
> ```
> $ python3
> Python 3.5.2 (default, Sep 10 2016, 08:21:44) 
> [GCC 5.4.0 20160609] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import tf2_py
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/home/nicu/dbot/overlay_ws/devel/lib/python2.7/dist-packages/tf2_py/__init__.py", line 35, in <module>
>     exec(__fh.read())
>   File "<string>", line 37, in <module>
> ImportError: No module named '_tf2'
> >>> 
> ```

Have you solved this ??? Coming from the future :D

Issue #175, opened 2016-05-11
cache.cpp introduces an ODR violation (with tf's cache.cpp)


macmason:
TF builds a cache.cpp, which defines (among other things) createExtrapolationException1. For some reason, these are defined in the global namespace.

TF2 defines the same functions, again in the global namespace. As TF depends on TF2, this is an ODR violation, so any call into these functions is undefined behavior.

macmason:
I can fix this by using a "namespace tf2 { ... }" rather than a "using namespace tf2" in cache.cpp. Presumably, we'll fallen afoul of some esoteric C++ name lookup rule.

tfoote:
Thanks for catching that. I've made it explicit in #176 does that work for you?

macmason:
Sure does. One nitpick on the change itself.

PR #176, opened 2016-05-11
Move internal cache functions into a namespace

tfoote:
Fixes https://github.com/ros/geometry2/issues/175

### DIFF
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -38,7 +38,7 @@
 #include <geometry_msgs/TransformStamped.h>
 #include <assert.h>
 
-using namespace tf2;
+namespace tf2 {
 
 TransformStorage::TransformStorage()
 {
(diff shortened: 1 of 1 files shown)


tf2/src/cache.cpp
```diff
@@ -60,6 +60,8 @@ TimeCache::TimeCache(ros::Duration max_storage_time)
 : max_storage_time_(max_storage_time)
 {}
```
macmason, line 62: I think you want to drop the "using namespace" at the top-of-file, too.

tfoote, line 62: I closed the namespace immediately below the declarations so the using tf2 is used through the rest of the file. Do you think it's cleaner to just put the tf2 namespace at the top instead of "using".

macmason, line 62: Yes, I think just enclosing everything in a "namespace foo { ... }" is cleaner. It mirrors the pattern in the header file.

macmason, line 62: namespace tf2 {
namespace cache {
...everything....
} 
}

is consistent with both the ROS C++ style guide and the Google C++ style guide, as well. So it's the way to go.

PR #179, opened 2016-06-01
Adds ability to load static TF from the ROS parameter server.

felixduvallet:
This PR adds the ability to pass a ROS parameter name to the tf publisher, where the parameter contains the TF to publish:

```
static_transform_publisher /param_name
```

This can simplify some launching behavior: the TF can be stored in a YAML file and loaded directly into the param server. One example use case is storing a calibration TF as a file, and publishing it using the static_transform_publisher.

I also made some minor simplifications to the code, if those are annoying let me know and I'll update the PR.

I tested this using Travis, the log is [here](https://travis-ci.org/felixduvallet/geometry2/builds/134489229).

### DIFF
--- a/test_tf2/package.xml
+++ b/test_tf2/package.xml
@@ -36,6 +36,7 @@
   <run_depend>tf2_msgs</run_depend>
 
   <test_depend>rosunit</test_depend>
+  <test_depend>rosbash</test_depend>
 
 </package>
 
--- a/test_tf2/test/static_publisher.launch
+++ b/test_tf2/test/static_publisher.launch
@@ -1,5 +1,33 @@
 <launch>
   <node pkg="tf2_ros" type="static_transform_publisher" name="test_static_pub_ab" args="1 0 0 0 0 0 1 a b" />
   <node pkg="tf2_ros" type="static_transform_publisher" name="test_static_pub_bc" args="0 1 0 0 0 0 1 b c" />
-  <test test-name="test_static_publisher" pkg="test_tf2" type="test_static_publisher" args="--text" />
+
+  <!-- Start the static tf publisher with a valid tf.  There is a test
+       (tf_from_param_server_valid) that ensures the TF is published. -->
+  <rosparam param="test_tf2/tf_valid"
+            file="$(find test_tf2)/test/test_tf_valid.yaml" />
+  <node pkg="tf2_ros" type="static_transform_publisher"
+        name="test_static_pub_param_valid"
+        args="test_tf2/tf_valid" />
+
+  <!-- Start the static tf publisher with an *invalid* tf.
+       The main purpose of this test is to ensure the node dies gracefully. -->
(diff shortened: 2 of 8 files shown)


tfoote:
Thanks this looks useful. Would you have time to make a basic unit test? A roslaunch with the embedded parameters and a test script which checks that the appropriate transform is published would be a good start. 

Thanks for running the unit tests of travis. We also have pull request building running on jenkins for this repo.

felixduvallet:
Good idea @tfoote , it was easy to add to the test_tf2 package.
I added three tests:
- start the publisher with a valid TF in the param server
- start the publisher with an invalid TF in the param server
- start the publisher with a non-existent parameter

The last two I wasn't quite sure what to assert other than graceful death... let me know if you have any ideas.

felixduvallet:
I'm also curious if you have any input as to the best place to document this feature so it doesn't remain 'hidden'.

tfoote:

tfoote:
For the negative tests, I'd suggest something like writing it in a separate test instance. And there can be a simple python test program that comes up and launches the static transform publisher and checks the return code is negative and it returns quickly. Unfortunately we don't have negative return code checking in rostest.

felixduvallet:
Thanks for the suggestion @tfoote . Just to be clear, are you suggesting I use the [roslaunch python API](http://wiki.ros.org/roslaunch/API%20Usage) to start the node and time how long it takes to die gracefully?

tfoote:
I was actually thinking of just using subprocess but you could use the roslaunch api as well. 

Also about documentation I'd suggest extending the documentation in [tf2_ros](http://wiki.ros.org/tf2_ros) which is included into the tutorials on static transforms [c++](http://wiki.ros.org/tf2/Tutorials/Writing%20a%20tf2%20static%20broadcaster%20%28C%2B%2B%29) [python](http://wiki.ros.org/tf2/Tutorials/Writing%20a%20tf2%20static%20broadcaster%20%28Python%29)

felixduvallet:
@tfoote please take another look.
I added unit tests that check the node exits (with the correct error code), and for the right reason (by checking the return message).
Note that this adds a test dependency on rosbash (necessary to be able to `rosrun` anything).

tfoote:

I'll try to review this fully tomorrow.

tfoote:
I'm rerunning the unit tests, it looks like there's a flaky unrelated test failure. But otherwise it looks good. 

I think we could work around the extra test dependency by working out the path to the executable, but test dependencies are not too expensive, and rosbash is pretty light weight so I'm not going to worry about it. 

Would you mind squashing the commits semantically? Otherwise I'll let github do a generic squash merge.

felixduvallet:
@tfoote okay, done (note: I've never done that before, if I messed up let me know!).

Let me know if you want me to squash further.

felixduvallet:

tfoote:
It looks good. Thanks for submitting this and iterating on it!

PR #188, opened 2016-09-17
tf2_eigen: eigenToTransform/transformToEigen has now a stamped and a non-stamped version.

cwecht:
Up to now there was only a quite strange version of eigenToTransform/transformToEigen in tf2_eigen: it provided a conversion between Eigen::Affine3d and geometry_msgs::TransformStamped. This is now solved more consistent: there are conversion functions between Eigen::Affine3d and geometry_msgs::Transform and between tf2::StampedEigen::Affine3d and geometry_msgs::TransformStamped.

### DIFF
--- a/tf2_eigen/include/tf2_eigen/tf2_eigen.h
+++ b/tf2_eigen/include/tf2_eigen/tf2_eigen.h
@@ -38,35 +38,57 @@
 namespace tf2
 {
 
-/** \brief Convert a timestamped transform to the equivalent Eigen data type.
- * \param t The transform to convert, as a geometry_msgs TransformedStamped message.
+/** \brief Convert a transform to the equivalent Eigen data type.
+ * \param t The transform to convert, as a geometry_msgs Transform message.
  * \return The transform message converted to an Eigen Affine3d transform.
  */
 inline
-Eigen::Affine3d transformToEigen(const geometry_msgs::TransformStamped& t) {
-  return Eigen::Affine3d(Eigen::Translation3d(t.transform.translation.x, t.transform.translation.y, t.transform.translation.z)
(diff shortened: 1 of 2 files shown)


tf2_eigen/include/tf2_eigen/tf2_eigen.h
```diff
@@ -38,38 +38,60 @@
...
  * \param t The transform to convert, as a geometry_msgs TransformedStamped message.
- * \return The transform message converted to an Eigen Affine3d transform.
+ * \return The transform message converted to an stamped Eigen Affine3d transform.
  */
 inline
-Eigen::Affine3d transformToEigen(const geometry_msgs::TransformStamped& t) {
-  return Eigen::Affine3d(Eigen::Translation3d(t.transform.translation.x, t.transform.translation.y, t.transform.translation.z)
-			 * Eigen::Quaterniond(t.transform.rotation.w,
-					      t.transform.rotation.x, t.transform.rotation.y, t.transform.rotation.z));
+tf2::Stamped<Eigen::Affine3d> transformToEigen(const geometry_msgs::TransformStamped& t) {
```
tfoote, line 56: If someone is using this we can't just change the return type. And unfortunately you can't overload on the return type. We'll need to keep this prototype but a new transformToEigenStamped method could be added with this implementation.

tfoote (changes requested):
This was implemented as a helper function for the methods below not so much as a public API, though since it's in the header and it's a bare function it's public. The recommendation is to use the [tf2::convert](http://wiki.ros.org/tf2/Tutorials/Migration/DataConversions) abstraction which then will convert to and from any datatype and support arbitrary input and output datatypes using templating.

PR #189, opened 2016-09-27
tf2_geometry_msgs: removed KDL dependencies

cwecht:
Instead of KDL, tf2 methods and types are now used to transform geometry_msgs.


tf2_geometry_msgs/setup.py
```diff
@@ -6,7 +6,7 @@
 d = generate_distutils_setup(
     packages=['tf2_geometry_msgs'],
     package_dir={'': 'src'},
-    requires={'rospy','geometry_msgs','tf2_ros','orocos_kdl'}
```
tfoote, line 9: There's no associated python code updates.

tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
```diff
@@ -186,73 +183,6 @@ void fromMsg(const geometry_msgs::PointStamped& msg, geometry_msgs::PointStamped
...
-/*****************/
-
-/** \brief Extract a timestamp from the header of a Pose message.
- * This function is a specialization of the getTimestamp template defined in tf2/convert.h.
- * \param t PoseStamped message to extract the timestamp from.
- * \return The timestamp of the message.
- */
-template <>
-inline
-  const ros::Time& getTimestamp(const geometry_msgs::PoseStamped& t)  {return t.header.stamp;}
```
tfoote, line 200: This is one of many places removing public API. They will all need to go through a deprecation cycle.

tfoote (changes requested):
Overall this is good idea to remove the dependency if possible. I have two comments though. Since this is a stable released package we cannot modify or remove API without a deprecation cycle and recommended migration path replacing the removed functions with equivalent ones elsewhere. And secondly this removes the dependency in the setup.py but does not change the python implementation to not use kdl.

tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
cwecht: The only change affecting the api is the removing gmTransformToKDL(); every other functions has just been moved in order to avoid some code duplication (git did not get this actually...). Of cause  gmTransformToKDL() can be re-added in order to do the deprecation cycle, if desired.

cwecht:
Removing KDL from the python implementation was not that easy,  because numpy does not provide a quaternion implementation, but I see basically three ways to do it:
1. using the python package [transforms3d](https://pypi.python.org/pypi/transforms3d). This obviously introduces a new dependency. If've  implemented a solution using this package [here](https://github.com/cwecht/geometry2/blob/tf2_geometry_msgs_replace_kdl_by_transform3d/tf2_geometry_msgs/src/tf2_geometry_msgs/tf2_geometry_msgs.py).
2. taking the relevant parts of transforms3ds quaternion implementation and putting it in `tf2_geometry_msgs.py`; this has been done [here](https://github.com/cwecht/geometry2/blob/tf2_geometry_msgs_replace_python_kdl/tf2_geometry_msgs/src/tf2_geometry_msgs/tf2_geometry_msgs.py).
3. create python-bindings to the c++-implementation, but I do not think, that it's worth the effort.

PR #191, opened 2016-11-04
Drop roslib.load_manifest

mikepurvis:
Unneeded with catkin, and it screws up doc-generation workflows that depend PYTHONPATH only (as opposed to ROS_PACKAGE_PATH).


tfoote:
Thanks

PR #192, opened 2016-11-28
Improve tf compatibility

roehling:
This pull request reimplements tf2_py API calls which are required to keep the new TF python proxy implementation fully backwards compatible.

See ros/geometry#134


tfoote (approved):
Thanks!

PR #193, opened 2016-12-01
made a fix for how rclcpp::spin was being used with std::thread

wjwwood:
Connects to ros2/rclcpp#277


tfoote (approved):
Seems fine to me. Too bad it can't infer that.

PR #196, opened 2017-01-04
fixes #194 check for quaternion normalization before inserting into storage

### LINKED ISSUE #194: check for quaternion normalization before inserting into storage
If this is not done. Invalid quaternions can cause future queries to return nan values. 
lookupTransform of Time(0) queries return the previously set invalid quaternions but not nans.

But time specific queries which use the interpolation give nans due to the interpolation trying to use the quaternions.

I expect that the most common case is all zero quaternions that have not been initialized correctly. 

It should be checked approximately here: https://github.com/ros/geometry2/blob/indigo-devel/tf2/src/buffer_core.cpp#L243

spaghetti-:
@tfoote Is this what you had in mind here at #194? 

This checks for normalization, but maybe a better idea is to check for `q.dot(q) != 0` for an invalid quaternion instead?

### DIFF
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -241,6 +241,19 @@ bool BufferCore::setTransform(const geometry_msgs::TransformStamped& transform_i
     error_exists = true;
   }
 
+  bool valid = std::abs((stripped.transform.rotation.w * stripped.transform.rotation.w
+                        + stripped.transform.rotation.x * stripped.transform.rotation.x
+                        + stripped.transform.rotation.y * stripped.transform.rotation.y
+                        + stripped.transform.rotation.z * stripped.transform.rotation.z) - 1.0f) < 10e-6;
+
+  if (!valid) 
+  {
+    logError("TF_DENORMALIZED_QUATERNION: Ignoring transform for child_frame_id \"%s\" from authority \"%s\" because of an invalid quaternion in the transform (%f %f %f %f)",
+             stripped.child_frame_id.c_str(), authority.c_str(),
+             stripped.transform.rotation.x, stripped.transform.rotation.y, stripped.transform.rotation.z, stripped.transform.rotation.w);
+    error_exists = true;
+  }
+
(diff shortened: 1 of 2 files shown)


tfoote:
Yeah, this is basically it. The test needs to have an epsilon of tolerance for floating point precision errors. I'm not sure what the tolerance should be. It looks like the errors go as a square of the error in normalization. https://stackoverflow.com/questions/11667783/quaternion-and-normalization We can assert the tolerance be pretty small since we're using doubles. 

And it would be great to add a test for setting some common invalid quaternions and asserting that it is rejected.

spaghetti-:
@tfoote changed to take tolerance, set it to 10e-6 if you think it needs to be smaller. It should also catch invalid quaternions.

Will add the tests soon.

spaghetti-:
There was an error in the logic which I fixed, separated it out into a boolean so that its more obvious, added tests. @tfoote

tfoote (approved):
Thanks for the iteration and tests.

tompe17:
I get the following:
`[ERROR] [1490090873.307973855]: Ignoring transform for child_frame_id "uav0/vicon_laser" from authority "unknown_publisher" because of an invalid quaternion in the transform (-0,018510 0,706860 -0,018510 0,706860)`

The tansform is generated by:
`<node  pkg="tf2_ros" type="static_transform_publisher" name="laser_frame_publisher" args="0.12 0 -0.1 -0.01851 0.70686 -0.01851 0.70686 /uav0/vicon /uav0/vicon_laser" />`

And I think I computed the values from the angles -90, -87. -90 and using C++ and printing it so I only got five decimals. This gives an error bigger than 10e-6.

I noticed it because old bag files did not display properly in rviz.  So I would suggest to increase the limit here.

This is also discussed here: http://answers.ros.org/question/257498/rviz-ignoring-transforms/

spaghetti-:
@tfoote We can either add an option to allow the tolerance to be changed or we can increase the tolerance.

PR #201, opened 2017-01-23
tf2_ros: add option to unregister TransformListener

achim-k:
This PR adds the option to unregister an existing `TransformListener` if it is not needed anymore. This is for example handy when looking up a static transform once during initialization of a rospy node.

Instead of listening to the tf topics in the `TransformListenerThread`, subscribers are now set up directly in the `TransformListener` class. To my knowledge, every rospy subscriber has its own thread, so the `TransformListenerThread` is actually not required (correct me if I'm wrong).

The new `unregister()` method simply unregisters the two tf subscribers.

### DIFF
--- a/tf2_ros/src/tf2_ros/transform_listener.py
+++ b/tf2_ros/src/tf2_ros/transform_listener.py
@@ -29,15 +29,12 @@
 
 import roslib; roslib.load_manifest('tf2_ros')
 import rospy
-import rospy
 import tf2_ros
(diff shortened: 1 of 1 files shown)


achim-k:
retest this please

tfoote:
You're right about the threading. The spin does not actually do anything except sleep in rospy. It looks like a big change but it does seem that it shouldn't actually change anything.

PR #202, opened 2017-01-30
tf2_geometry_msgs: added missing conversion functions

nlamprian:
Fixed a few indentation and documentation errors, and added the missing fromMsg and toMsg functions.


vincentrou:
It seems interesting to have this PR integrated which adds conversion to tf2::Vector3

PR #204, opened 2017-03-01
additional changes

Karsten1987:
if you don't mind, I opened a PR against your PR for adding my changes as well.

### DIFF
--- a/tf2_ros/include/tf2_ros/static_transform_broadcaster.h
+++ b/tf2_ros/include/tf2_ros/static_transform_broadcaster.h
@@ -49,8 +49,6 @@ namespace tf2_ros
 
 class StaticTransformBroadcaster{
 public:
-  /** \brief Constructor (needs a ros::Node reference) */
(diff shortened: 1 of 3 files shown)


tf2_ros/include/tf2_ros/static_transform_broadcaster.h
```diff
@@ -52,6 +52,8 @@ class StaticTransformBroadcaster{
   /** \brief Constructor (needs a ros::Node reference) */
   StaticTransformBroadcaster();
```
gerkey: I think that this constructor prototype should be removed (or else implemented with internal creation of a node, but I think that it's clearer to force the user to pass in a node as is done with the other constructor).

Karsten1987: Agreed. Should have been removed.
Updated the PR.

gerkey:
thanks!

PR #209, opened 2017-03-15
BufferServers uses node name as action namespace

jihoonl:
instead of fixed namespace `tf2_buffer_server`.

 It confused me when I tried to use two tf2 buffer server with different names.


tfoote (changes requested):
This looks like a general improvement. However this would break the exposed API within the release cycle. To land this change we would need to provide backwards compatibility so we wouldn't break all launch files using this node.

tfoote:
replaced by #296

PR #211, opened 2017-03-20
BUFFER_CORE: Add erase function.

lsouchet:
This includes:
- structure containers change to boost::flat_map to dynamically handle
  data access after a deletion.
- erase() function adding to clear the data from a frame entry

### DIFF
--- a/tf2/include/tf2/buffer_core.h
+++ b/tf2/include/tf2/buffer_core.h
@@ -46,6 +46,7 @@
 //////////////////////////backwards startup for porting
 //#include "tf/tf.h"
 
+#include <boost/container/flat_map.hpp>
 #include <boost/unordered_map.hpp>
 #include <boost/thread/mutex.hpp>
 #include <boost/function.hpp>
@@ -102,6 +103,12 @@ public:
 
   /** \brief Clear all data */
   void clear();
+
+  /** \brief Erase a specified frame
+   * \param frame_id_str The name of the frame that will be deleted.
+   * This function will remove the frame, and reparent all its children frames to NO_PARENT.
+   */
+  void erase(const std::string& frame_id_str);
 
   /** \brief Add transform information to the tf data structure
    * \param transform The transform to store
@@ -317,9 +324,9 @@ private:
   
   /** \brief The pointers to potential frames that the tree can be made of.
    * The frames will be dynamically allocated at run time when set the first time. */
-  typedef std::vector<TimeCacheInterfacePtr> V_TimeCacheInterface;
-  V_TimeCacheInterface frames_;
-  
+  typedef boost::container::flat_map<CompactFrameID, TimeCacheInterfacePtr> M_TimeCacheInterface;
+  M_TimeCacheInterface frames_;
+
   /** \brief A mutex to protect testing and allocating new frames on the above vector. */
   mutable boost::mutex frame_mutex_;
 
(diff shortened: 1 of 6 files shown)


tfoote:
Can you clarify what you intend this new API to do and what the use case is that it's serving?

lsouchet:
The purpose of this new API is to free the memory that is not used anymore because the frames are not used. 
In my use case, I store frames shared_ptrs that are linked to a frame in the BufferCore. When their refcount reaches 0 because nobody is using them anymore, I would like to free them to avoid my bufferCore to basically leaking.
As I may create many "temporary" frames, it may result in lots of unused frames in memory.

This PR thus adds an erase() function to erase a frame from the BufferCore, and to do so refactors a bit the data structures used in the BufferCore to match this goal.

sbarthelemy:
@tfoote 

The goal is to have a way to erase the frames that are not needed anymore.
It frees memory.

This is especially useful if applications create temporary frames and uptimes are high.

tfoote:
Yeah, if there are a lot of temporary frames I can see that they might leak. However I'm not sure that the proposed 'erase' method is the best approach to this. This both breaks the premise of the cache which could potentially side effect other users of the Buffer, as well as requires the programmer to specifically track which frame_ids they want to clear.

I would suggest a simpler method which simply clears old data from the cache. As an optimization each linked list is only pruned on insert with respect to the cache duration. However it could easily be pruned at any time. A simple method cleanup cache could call the prune method for all frames given the current time. This would keep cache consistency during the valid period, but would clear old temporary transforms from memory. Adding this behavior as an automated periodic behavior that could be enabled might be worth considering too.

On the technical side. Looking forward (aka ROS2) we're moving away from using boost at all. And so restructuring to use a new boost datatype is not something that we want to do either.

lsouchet:
@tfoote 
Thanks for your answer, but I believe you suggestion does not quite match our need, as it will not free all the frames memory but only the transform data leaving dangling frames in memory. 
erase() method is standard for a c++ container, and indeed requires programmers to track element lifetime. In our application, we are doing that using shared_ptrs above the BufferCore API.

To give you some figures about our use case, some client applications of our API can for example publish frames related to people detection up to 10 times per second, leading to a leak of a few decades of bytes each time, and a few megabytes per day. For a robot whose target uptime is in weeks, this is too much.

Regarding the boost container issue, the flat_map is knowingly more efficient for this kind of application but if it appears to be a blocking point, we might consider using std::unordered_map or some similar structure.

tfoote:
Are you actually seeing megabytes accumulating per day? If you're publishing 10 messages per second you should have a maximum of approximately 100 messages per frame id since the Cache discards old data once it is older than the cache duration, which defaults to 10 seconds. Thus you don't get continuous accumulation in long running applications.

For a long running application as I mentioned above the cache invalidation is done on a per frame_id basis. This could be generalized to clear the cache across all frame_ids periodically to clear previously published frames from the cache. With this improvement you will never have more than 100 messages per active frame_id and once a frame_id goes inactive the majority of it's data will be cleared within 10 seconds. 

The BufferCore is not designed as a generic long term storage container. If you want to be doing that sort of thing there are a lot of optimizations that should be made rather than just adding an erase method to the core. Including decimating duplicate data, storing the data in queryable cointainer instead of a list. The choice of a list is to optimize data recording and from benchmarking is eqivalent in streaming data use cases. This is also why the Buffer is passed into the Listener and other APIs to allow you to provide a different implementation optimized for your use case.

lsouchet:
@tfoote 
> Are you actually seeing megabytes accumulating per day?

Yes. Not because of new messages, but because of new **frames** allocated [here](https://github.com/ros/geometry2/blob/2ac6e235dbcbd83704aacf989314dbffb9c7cc20/tf2/src/buffer_core.cpp#L290) and [there](https://github.com/ros/geometry2/blob/2ac6e235dbcbd83704aacf989314dbffb9c7cc20/tf2/src/buffer_core.cpp#L292) and never freed until BufferCore destruction.

Decimating has nothing to do with our problem, we are not concerned by clearing frame buffers but rather by clearing frames themselves. By the way we kept the default buffer duration and it is fine with our usage. Adding an erase() is necessary and sufficient to solve our problem.
Maybe you haven't encountered this problem already because it has not been used in application creating many frames (and not only new messages).

tfoote:
Can you clarify what scale of "many" you are talking about when you say "many frames"? Empty caches should use on the order of 10s of bytes each. The empty caches either have one Transform stored inside them or a linked list of transforms.

But either way I would still not want to try to support arbitrary modifications of the active buffer that could potentially side effect other lookups inside the valid cache time. My above suggestion of a method that clears old data could also deallocate the (Static)TimeCache(s) if they are found to be empty after pruning.

lsouchet:
Alright, I am not sure about what you mean by the "arbitrary modifications" but I understand that you are not comfortable with merging this PR.
Efficiently managing memory is too important in our use case, so we will find a way around this issue.

Thanks for your time anyway! :)

PR #212, opened 2017-03-22
fix installation dir on windows

Karsten1987:
fixes the installation path under windows to be on the `PATH`


wjwwood (approved):
lgtm, since `tf2` is a library.

PR #213, opened 2017-03-22
Add tf2::TimePoint <-> builtin_interfaces::msg::Time conversions

dhood:
example usage: `tf2_ros::fromMsg(t.header.stamp)`


dhood:
oops this is in the wrong place. I will re-open on ros2 fork

PR #215, opened 2017-03-24
Remove an unused variable from the tf2_eigen test.

clalancette:
While we are in here, also kill off some trailing whitespace.

### DIFF
--- a/tf2_eigen/test/tf2_eigen-test.cpp
+++ b/tf2_eigen/test/tf2_eigen-test.cpp
@@ -32,8 +32,6 @@
 #include <gtest/gtest.h>
 #include <tf2/convert.h>
 
-static const double EPS = 1e-3;
-
 TEST(TfEigen, ConvertVector3dStamped)
 {
   const tf2::Stamped<Eigen::Vector3d> v(Eigen::Vector3d(1,2,3), ros::Time(5), "test");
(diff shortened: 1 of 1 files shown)


clalancette:
The kinetic build is "unstable", but in the same way that all previous builds are; it looks like it mostly comes from the eigen compatibility stuff for older distributions.  So I don't think we need to worry about that one w.r.t. this pull request.

The jade one also is "unstable", for the same reason as the kinetic one.  The odd thing is that all the previous (and subsequent) builds are marked green, while just this one is flagged for being unstable.  Not sure what is going on there.

mikaelarguedas:
The jobs are marked as unstable if there are compiler warning only since kinetic (so not for Jade and Indigo).
The jade failure is unrelated to the warning, it's due to a test failure

clalancette:
Oh, right, now I see.  I must have clicked on the wrong thing before.  Thanks for the heads up.  That failing test looks unrelated to this PR; tf2_ros doesn't depend on tf2_eigen in any way, and I didn't touch any other code that would change that (as far as I can tell).

mikaelarguedas:
Yes it looks like a flaky test based on the job history. The compiler warning should be fixed though because otherwise the maintainer will receive an email for every PR opened and commit added to this repo

clalancette:
Yeah, I agree.  @k-okada ended up doing a fairly nice solution for eigen_stl_containers; I'll look at porting that over to tf2_eigen.

In any case, that problem is independent of this PR, so we can address it separately.

tfoote:

tfoote:

PR #216, opened 2017-03-24
Switch tf2_eigen to use package.xml format 2.

clalancette:
This is to modernize it a bit to make it more compatible
with ROS 2.  While we are in here, remove the run/exec dependency
on cmake_modules, which doesn't seem to be needed at runtime.

### DIFF
--- a/tf2_eigen/package.xml
+++ b/tf2_eigen/package.xml
@@ -9,12 +9,12 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
+  <depend>geometry_msgs</depend>
+  <depend>tf2</depend>
+
   <build_depend>cmake_modules</build_depend>
   <build_depend>eigen</build_depend>
-  <build_depend>geometry_msgs</build_depend>
-  <build_depend>tf2</build_depend>
 
-  <exec_depend>geometry_msgs</exec_depend>
-  <exec_depend>tf2</exec_depend>
+  <build_export_depend>eigen</build_export_depend>
 
 </package>
--- a/tf2_eigen/package.xml
+++ b/tf2_eigen/package.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<package>
+<package format="2">
   <name>tf2_eigen</name>
   <version>0.5.15</version>
   <description>tf2_eigen</description>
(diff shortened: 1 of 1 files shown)


tf2_eigen/package.xml
```diff
@@ -8,13 +8,13 @@
...
+  <build_depend>cmake_modules</build_depend>
   <build_depend>eigen</build_depend>
   <build_depend>geometry_msgs</build_depend>
   <build_depend>tf2</build_depend>
-  <build_depend>cmake_modules</build_depend>
 
-  <run_depend>geometry_msgs</run_depend>
-  <run_depend>tf2</run_depend>
-  <run_depend>cmake_modules</run_depend>
+  <exec_depend>geometry_msgs</exec_depend>
```
tfoote: geometry_msgs and tf2 also needs the `<build_export_depend>` because of the includes in the headers. http://docs.ros.org/kinetic/api/catkin/html/howto/format2/migrating_from_format_1.html#run-depend

But since they overlap with the build depends it can be collapsed to just a `<depend>` .

tf2_eigen/package.xml
```diff
@@ -8,13 +8,13 @@
   <license>BSD</license>
 
   <buildtool_depend>catkin</buildtool_depend>
+
+  <build_depend>cmake_modules</build_depend>
```
tfoote: This needs `<build_export_depend>` too since it's in the header and will be needed downstream.

tfoote (changes requested):
Thanks for the update. I see two small tweaks needed to make sure to export the build depends.

tf2_eigen/package.xml
clalancette: Can you explain this a bit further so I understand?  From what I can tell, tf2_eigen.h includes:

#include <tf2/convert.h>
#include <Eigen/Geometry>
#include <geometry_msgs/PointStamped.h>
#include <geometry_msgs/PoseStamped.h>

So all of those would need build_export_depend (as you point out below).  Why would we also need to build_export_depend cmake_modules?

tf2_eigen/package.xml
clalancette: Right, good point.  Will fix that.

tf2_eigen/package.xml
tfoote: I think you're right. All the EIGEN variables will be evaluated so it won't need them downstream.

tfoote (approved):

clalancette:
Thanks :).

PR #217, opened 2017-03-28
Switch tf2_geometry_msgs to use package.xml format 2

dhood:


tf2_geometry_msgs/package.xml
```diff
@@ -10,19 +10,16 @@
 
   <url type="website">http://www.ros.org/wiki/tf2_ros</url>
     
-  <buildtool_depend version_gte="0.5.68">catkin</buildtool_depend>
+  <buildtool_depend>catkin</buildtool_depend>
```
mikaelarguedas: Does this package now compiles with catkin older than 0.5.68 ?

PR #220, opened 2017-04-01
relax normalization tolerance. #196 was too strict for some use cases.

tfoote:
Follow up #196 

@spaghetti- @tompe17

### DIFF
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -41,6 +41,9 @@
 namespace tf2
 {
 
+// Tolerance for acceptable quaternion normalization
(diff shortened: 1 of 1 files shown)


spaghetti-:
I personally think its a bit too relaxed? Looks good otherwise.

tfoote:
@spaghetti- I agree that I'd like it to have tighter tolerances, but there's a level that the level of accuracy is something that the deployment can choose. From my reading the tolerances are completely use case dependent and the computational errors grow as a square of the non-normalization. However if the measurement itself has high uncertainty enforcing super high precision computation does not effectively improve the system performance, and will break some current use cases.

spaghetti-:
I see, your observation about the uncertainty being a bigger problem is quite true. makes sense. lgtm then.

Issue #219, opened 2017-03-30
Strict type checks (#197) on python interface of Buffer::setTransform make direct input from bagfile cumbersome


stwirth:
#197 introduced new type checks on the python interface of `Buffer::setTransform` to fix #159.
Unfortunately this makes filling a tf buffer from a bagfile cumbersome, as bagfiles have their own type definitions for messages.
Here is an example program that demonstrates the problem:
```
import tf2_ros
import rosbag
from geometry_msgs.msg import TransformStamped, Vector3, Quaternion

def set_transform_test(t):
    print "setting transform of type {}".format(type(t))
    tf2_buf = tf2_ros.Buffer(None, False)
    tf2_buf.set_transform(t, 'default_authority')

if __name__ == '__main__':
    # create dummy transform
    t = TransformStamped()
    t.header.frame_id = 'parent'
    t.child_frame_id = 'child'
    t.transform.translation = Vector3()
    t.transform.rotation = Quaternion(0, 0, 0, 1)

    set_transform_test(t)

    # write transform to bag and read it again
    wbag = rosbag.Bag('test.bag', 'w')
    wbag.write('transform', t)
    wbag.close()

    rbag = rosbag.Bag('test.bag', 'r')
    for topic, msg, t in rbag.read_messages(topics=['transform']):
        set_transform_test(msg)
    rbag.close()
```
This is the output:
```
$ python tf_bag_test.py 
setting transform of type <class 'geometry_msgs.msg._TransformStamped.TransformStamped'>
setting transform of type <class 'tmpTmLCns._geometry_msgs__TransformStamped'>
Traceback (most recent call last):
  File "tf_bag_test.py", line 28, in <module>
    set_transform_test(msg)
  File "tf_bag_test.py", line 9, in set_transform_test
    tf2_buf.set_transform(t, 'default_authority')
TypeError: transform.translation must be of type Vector3
```
To work around this you need to manually transform the type `tmpTmLCns._geometry_msgs__TransformStamped` to type `geometry_msgs.msg._TransformStamped.TransformStamped`.

I would propose to make the type checks less strict, e.g. just enforcing that the transform has a member `translation` which has three floating point values and a member `rotation` that has four.

/cc @spaghetti-

spaghetti-:
Good catch, I think that enforcing members `translation` and `rotation` is good enough as well. I'll try to send in a PR this week.

stwirth:
awesome!

spaghetti-:
[...]

PR #221, opened 2017-04-03
Relax strict type checks at setTransform to only check for members

spaghetti-:
Previously introduced type check at the buffer::setTransform python
interface checks the supplied variable against the strict
geometry_msgs::Vector3 and geometry_msgs::Quaternion types[0]. This was
later reported to make input from bag files cumbersome because of
temporary intermediate types[1].

This relaxes the type check and only ensures that the expected members
are present. It still issues a UserWarning about the type mismatch.

[0] PR #197
[1] Issue #219

### DIFF
--- a/tf2_py/src/tf2_py.cpp
+++ b/tf2_py/src/tf2_py.cpp
@@ -377,13 +377,28 @@ static PyObject *lookupTwistFullCore(PyObject *self, PyObject *args)
 static inline int checkTranslationType(PyObject* o)
 {
   PyTypeObject *translation_type = (PyTypeObject*) PyObject_GetAttrString(pModulegeometrymsgs, "Vector3");
-  return PyObject_TypeCheck(o, translation_type);
+  int type_check = PyObject_TypeCheck(o, translation_type);
+  int attr_check = PyObject_HasAttrString(o, "x") &&
+                   PyObject_HasAttrString(o, "y") &&
+                   PyObject_HasAttrString(o, "z");
+  if (!type_check) {
+    PyErr_WarnEx(PyExc_UserWarning, "translation should be of type Vector3", 1);
+  }
+  return attr_check;
 }
 
 static inline int checkRotationType(PyObject* o)
 {
   PyTypeObject *rotation_type = (PyTypeObject*) PyObject_GetAttrString(pModulegeometrymsgs, "Quaternion");
-  return PyObject_TypeCheck(o, rotation_type);
+  int type_check = PyObject_TypeCheck(o, rotation_type);
+  int attr_check = PyObject_HasAttrString(o, "w") &&
+                   PyObject_HasAttrString(o, "x") &&
+                   PyObject_HasAttrString(o, "y") &&
+                   PyObject_HasAttrString(o, "z");
+  if (!type_check) {
+    PyErr_WarnEx(PyExc_UserWarning, "translation should be of type Quaternion", 1);
+  }
+  return attr_check;
 }
 
 static PyObject *setTransform(PyObject *self, PyObject *args)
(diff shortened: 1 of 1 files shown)


stwirth:
Thanks for the quick fix!
+1 for merging!

tf2_py/src/tf2_py.cpp
```diff
@@ -377,13 +377,28 @@ static PyObject *lookupTwistFullCore(PyObject *self, PyObject *args)
 static inline int checkTranslationType(PyObject* o)
 {
   PyTypeObject *translation_type = (PyTypeObject*) PyObject_GetAttrString(pModulegeometrymsgs, "Vector3");
-  return PyObject_TypeCheck(o, translation_type);
+  int type_check = PyObject_TypeCheck(o, translation_type);
+  int attr_check = PyObject_HasAttrString(o, "x") &&
+                   PyObject_HasAttrString(o, "y") &&
+                   PyObject_HasAttrString(o, "z");
+  if (!type_check) {
+    PyErr_WarnEx(PyExc_UserWarning, "translation should be of type Vector3", 1);
```
tfoote, line 385: Duck typing is a little bit more pythonic. It seems that this warning would be thrown in known valid cases, and there wouldn't be a way to suppress it thus maybe it shouldn't be  a warning if the duck typing passes?

spaghetti-: I suppose I can take out the strict type check entirely then. Maybe I'll enforce that the members are also floats/ints etc?

tfoote: Yeah, checking the type would be good too.

tfoote:

tfoote:
@spaghetti- Thanks for iterating on this and @stwirth for the feedback.

PeterQFR:
running into this problem too,  how would I get this fix downstream? It seems my ros packages are up to date for kinetic.
In the meantime, a hack is to rebuild the Transform stamped type from the ground up after it is read from the bag file.

tf2_py/src/tf2_py.cpp
```diff
@@ -377,13 +377,28 @@ static PyObject *lookupTwistFullCore(PyObject *self, PyObject *args)
...
 {
   PyTypeObject *rotation_type = (PyTypeObject*) PyObject_GetAttrString(pModulegeometrymsgs, "Quaternion");
-  return PyObject_TypeCheck(o, rotation_type);
+  int type_check = PyObject_TypeCheck(o, rotation_type);
+  int attr_check = PyObject_HasAttrString(o, "w") &&
+                   PyObject_HasAttrString(o, "x") &&
+                   PyObject_HasAttrString(o, "y") &&
+                   PyObject_HasAttrString(o, "z");
+  if (!type_check) {
+    PyErr_WarnEx(PyExc_UserWarning, "translation should be of type Quaternion", 1);
```
darkdragon-001, line 399: The text should state _rotation_ instead of _translation_!

tf2_py/src/tf2_py.cpp
jaredkhan: This strict type checking didn't get removed in the end. Don't see a reason for this so would you be open to a PR that gets rid of this type check?

tfoote: If you'd like to improve it that would be appreciated. I'd suggest doing the type check first whickly. And if that doesn't pass do the duck typing adding checks for the type as well as the attribute name.

jaredkhan: Hmm. Perhaps but I think if passing a Quaternion is going to be much quicker than passing something that quacks like a Quaternion then the user probably _should_ be made aware of that. 

Good point about checking the type of the attributes. Looks like at the moment if you pass something with an 'x' attribute that happens not to be convertible to a double, it gets the value -1.0 thanks to PyFloat_AsDouble

PR #223, opened 2017-04-05
tf2_geometry_msgs: using tf2::Transform in doTransform-functions, mar…

cwecht:
…ked gmTransformToKDL as deprecated


cwecht:
This is a cleaner version of #189 with less changes. `gmTransformToKDL` has been marked deprecated as suggested in the [ROS DeveloperGuidelines](http://wiki.ros.org/DevelopersGuide#Deprecation). No other changes have been applied to the existing external interface.

PR #225, opened 2017-04-19
Document lifetime of returned reference for getTimestamp and getFrameId

dhood:
This just updates documentation.

For efforts porting tf2 to ROS 2 in https://github.com/ros2/geometry2/pull/14, we need to update the signature to not return const references. To reduce divergence, it's preferable for the signatures to be updated in ROS 1 as well, but I think the conclusion that @tfoote and I came to in our discussion was to leave it as-is for compatibility with external implementations of `tf/convert.h`


tfoote:

PR #227, opened 2017-04-25
Remove generate_rand_vectors() from a number of tests.

clalancette:
It was never used, so there is no reason to carry it around.
While I was in these tests, kill a bunch of trailing
whitespace.


tfoote (changes requested):
Thanks these are definitely removable. However your commit is majority whitespace changes which are unrelated and may cause issues with the current outstanding PRs. We can consider cleaning it up, but we should at least isolate it into a separate commit to allow easier merging.

clalancette:
Yeah, that's fine.  I'll split this into 2 different commits, one for the whitespace, one for the change.

PR #230, opened 2017-05-06
tf2_eigen: added support for Quaternion and QuaternionStamped

cwecht:


tfoote (approved):
Thanks. This looks good to me.

PR #234, opened 2017-05-09
Log jump duration on backwards time jump detection

stwirth:
TransformListener prints out a warning when ROS time jumps backwards.
This commit modifies the log line to include the amount of time that the jump had.


tfoote:
Thanks this looks helpful.

PR #235, opened 2017-05-10
Copy transform before altering it in do_transform_vector3 [issue 233]

### LINKED ISSUE #233: do_transform_vector3() alters the passed transformation
I'm in a situation where i use a transformation to transform several objects pose, velocities and accelerations. The latter two are transformed by do_transform_vector3 as they must not be translated but only rotated. The problem is when looping over the objects, the fist call of do_transform_vector3 sets the translation parameters of the transformation to zero (see `tf2_geometry_msgs/tf2_geometry_msgs.py` lines 71 -73), making all upcoming calls of do_transform_pose using a wrong transformation. It can be quickly solved by wrapping the transformation in the function call with a `copy.deepcopy()` statement.

I'm not exactly sure if this can be regarded as an issue, but i do not think it is intuitive to have a transformation call alter my inputs.

swagner91:
Fixes #233, where a transform passed to do_transform_vector3 was changes inside the function and made useless for later transforms.

### DIFF
--- a/tf2_geometry_msgs/src/tf2_geometry_msgs/tf2_geometry_msgs.py
+++ b/tf2_geometry_msgs/src/tf2_geometry_msgs/tf2_geometry_msgs.py
@@ -31,6 +31,7 @@ from geometry_msgs.msg import PoseStamped, Vector3Stamped, PointStamped
 import PyKDL
 import rospy
 import tf2_ros
+import copy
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Thanks this looks good.

tfoote:

PR #237, opened 2017-06-14
fix return value to prevent warnings on windows

mikaelarguedas:
Windows complain if we return a bool instead of an int here


dhood:
see also https://github.com/ros/geometry2/pull/229/files

tfoote (approved):

PR #241, opened 2017-06-22
Fix do_transform_cloud for multi-channelled pointcloud2.

at-wat:
When multi-channelled pointcloud2 (e.g. index channel) is passed to
do_transform_cloud, pack_into(buff, offset, *p) returns

    error: pack_into expected 4 items for packing (got 3)

since it tries to pack only xyz data with original field list.

This commit appends auxiliary channel data to output data.


at-wat:
This should realize same behavior to the cpp implementation of tf2::doTransform.

tf2::doTransform first copies whole PointCloud2. (https://github.com/ros/geometry2/blob/indigo-devel/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.h#L61)
So, all of the auxiliary channel data is kept in output on its cpp implementation.

tf2_sensor_msgs/test/test_tf2_sensor_msgs.py
```diff
@@ -45,7 +45,48 @@ def test_simple_transform(self):
...
+        # we add two points (with x, y, z to the cloud)
+        self.point_cloud_in.width = 2
+        self.point_cloud_in.row_step = self.point_cloud_in.point_step * self.point_cloud_in.width
+
+        points = [1, 2, 0, 123, 10, 20, 30, 456]
+        self.point_cloud_in.data = struct.pack('3fi3fi', *points)
+
+
+        self.transform_translate_xyz_300 = TransformStamped()
+        self.transform_translate_xyz_300.transform.translation.x = 300
```
tfoote: It would be great to reuse k from https://github.com/ros/geometry2/pull/241/files#diff-1eb6abfaf9a31338a0944f59de492214R80 maybe make that a constant at the top `TRANSFORM_OFFSET_DISTANCE = 300`

tf2_sensor_msgs/test/test_tf2_sensor_msgs.py
```diff
@@ -45,7 +45,48 @@ def test_simple_transform(self):
...
+        self.transform_translate_xyz_300.transform.translation.y = 300
+        self.transform_translate_xyz_300.transform.translation.z = 300
+        self.transform_translate_xyz_300.transform.rotation.w = 1  # no rotation so we only set w
+
+        assert(list(point_cloud2.read_points(self.point_cloud_in)) == [(1.0, 2.0, 0.0, 123), (10.0, 20.0, 30.0, 456)])
```
tfoote: It would be good to reuse the points defined on line 64 instead of duplicating the values.

tf2_sensor_msgs/test/test_tf2_sensor_msgs.py
```diff
@@ -45,7 +45,48 @@ def test_simple_transform(self):
...
+
+    def test_simple_transform_multichannel(self):
+        old_data = copy.deepcopy(self.point_cloud_in.data)  # deepcopy is not required here because we have a str for now
+        point_cloud_transformed = tf2_sensor_msgs.do_transform_cloud(self.point_cloud_in, self.transform_translate_xyz_300)
+
+        k = 300
+        expected_coordinates = [(1+k, 2+k, 0+k, 123), (10+k, 20+k, 30+k, 456)] # index channel (123, 456) must be kept same
```
tfoote: Same here, reusing the points from line 64 would be great.

tfoote (changes requested):
Thanks for the contribution. This looks good. I had two comments about maintainability of the tests.
If you store the offset points in the class you can reuse them to test for the result instead of hard coding them in the checks.

at-wat:
I haven't touched the original test without auxiliary channels, that have same hard codes, to keep the commit atomic.

tfoote (approved):
Thanks for the revision. It looks good.

> I haven't touched the original test without auxiliary channels, that have same hard codes, to keep the commit atomic.

Thanks that's appreciated. If you have a moment that clean up would be great to do there too.

Issue #248, opened 2017-07-07
Cannot change subscriber queue in transform_listener 


kojiterada:
Hi.

If tf is operated on computation resources with high load and cheap computing resources, problems due to over queue are occurring.
My suggestion is to be able to set the subscriber's queue.

tfoote:
resolved in #249

PR #249, opened 2017-07-07
Make the subscriber queue size mutable.

kojiterada:
Make the subscriber queue size mutable. #248

### DIFF
--- a/tf2_ros/src/tf2_ros/transform_listener.py
+++ b/tf2_ros/src/tf2_ros/transform_listener.py
@@ -38,7 +38,7 @@ class TransformListener():
     This class takes an object that instantiates the :class:`BufferInterface` interface, to which
     it propagates changes to the tf frame graph.
     """
-    def __init__(self, buffer):
+    def __init__(self, buffer, queue_size=None):
         """
         .. function:: __init__(buffer)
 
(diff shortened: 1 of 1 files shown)


tf2_ros/src/tf2_ros/transform_listener.py
```diff
@@ -47,9 +47,9 @@ def __init__(self, buffer):
             :param buffer: The buffer to propagate changes to when tf info updates.
         """
         self.buffer = buffer
-        self.tf_sub = rospy.Subscriber("/tf", TFMessage, self.callback)
+        self.tf_sub = rospy.Subscriber("/tf", TFMessage, self.callback, queue_size=queue_size)
         self.tf_static_sub = rospy.Subscriber("/tf_static", TFMessage, self.static_callback)
```
tfoote: The queue_size should also propogate here.

tf2_ros/src/tf2_ros/transform_listener.py
```diff
@@ -47,9 +47,9 @@ def __init__(self, buffer):
             :param buffer: The buffer to propagate changes to when tf info updates.
```
tfoote, line 47: Please update the doc block too.

tfoote (changes requested):
Thanks this looks like a good extension. I had a few comments. 

And when updating the doc block please also comment about side effecting on other subscribers: http://docs.ros.org/kinetic/api/rospy/html/rospy.topics.Subscriber-class.html wrt to the queue size.

tf2_ros/src/tf2_ros/transform_listener.py
kojiterada: I updated doc block.

tf2_ros/src/tf2_ros/transform_listener.py
kojiterada: I also change there.

tfoote (approved):
Thanks for the iteration. And exposing the extra options looks good too.

PR #251, opened 2017-07-12
fix memory leak calling Py_DECREF for all created PyObject

sergioramosjubierre:
I have found a memory leak in tf2_py.cpp

In the transform_converter function the PyObject objects have the Py_DECREF in order to decrease the Reference Counts and allow the object to be deallocated.
For example:
PyObject *ptranslation = PyObject_GetAttrString(ptransform, "translation");
...
Py_DECREF(ptranslation);

But not all of them. We can find things like this:
PyObject_SetAttrString(ptranslation, "x", PyFloat_FromDouble(transform->transform.translation.x));
PyFloat_FromDouble creates a new object, but PyObject_SetAttrString doesn't transfer the reference to ptranslation. That makes that the reference to the PyObject created by PyFloat_FromDouble is not decreased and then never deallocated.

There is information about this here:
https://docs.python.org/3/c-api/intro.html#reference-count-details
And also a discussion about this specific thing:
https://www.thecodingforums.com/threads/python-c-api-and-references.705035/

In this commit I have called explicitly Py_DECREF for all the created PyObject.
Whit this fix the memory leak is fixed.

We have experience this memory leak when having robots working during days.
So please, when this fix is merged I would like to ask for a release.

### DIFF
--- a/tf2_py/src/tf2_py.cpp
+++ b/tf2_py/src/tf2_py.cpp
@@ -105,7 +105,9 @@ static PyObject *transform_converter(const geometry_msgs::TransformStamped* tran
   PyObject_SetAttrString(pheader, "stamp", time_obj);
   Py_DECREF(time_obj);
 
-  PyObject_SetAttrString(pheader, "frame_id", stringToPython(transform->header.frame_id));
+  PyObject *frame_id = stringToPython(transform->header.frame_id);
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks. That looks like a pretty clear upgrade based on the docs. I'm triaging tickets this week and look to make a release before then end of the week.

PR #254, opened 2017-07-13
Apply 202 and 223 together

tfoote:
Since #202 and #223 had a large amount over overlap, this is an update that merges both. 

@cwecht @nlamprian if you have a chance to review that would be great. I believe that I've captured the net results of both your changes. And leveraged the additional conversions to make the transform methods simpler too.

I'm going to move onto triaging geometry and come back here before making a coordinated release.

### DIFF
--- a/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
+++ b/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
@@ -49,13 +49,29 @@ namespace tf2
 /** \brief Convert a TransformStamped message to a KDL frame.
  * \param t TransformStamped message to convert.
  * \return The converted KDL Frame.
+ * \deprecated
  */
+inline
+KDL::Frame gmTransformToKDL(const geometry_msgs::TransformStamped& t) __attribute__ ((deprecated));
 inline
 KDL::Frame gmTransformToKDL(const geometry_msgs::TransformStamped& t)
   {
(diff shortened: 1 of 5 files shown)


tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
```diff
@@ -492,41 +526,167 @@ geometry_msgs::Transform toMsg(const tf2::Transform& in)
...
+void doTransform(const geometry_msgs::TransformStamped& t_in, geometry_msgs::TransformStamped& t_out, const geometry_msgs::TransformStamped& transform)
+  {
+    tf2::Transform input;
+    fromMsg(t_in.transform, input);
+
+    tf2::Transform t;
+    fromMsg(transform.transform, t);
+    tf2::Transform v_out = t * input;
+
+    toMsg(v_out);
```
cwecht: @tfoote something is missing here :)

cwecht:
Unit-test for doTransform with Transforms and Quaternions would be useful.

cwecht:
I've done a unit-test for TransformStamped [here](https://github.com/cwecht/geometry2/commit/bd02acd434ab99e557dc2c939edf7a397a41c6d9) (which has been a low hanging fruit). An comparable test for quaternions would either trivial or require more effort (and would be rather unreadable).

nuclearsandwich:


edit: jk it merged :grin:

PR #257, opened 2017-07-16
Make tf2_py python3 compatible again

de-vri-es:
120a532 reintroduced usage of `PyString_FromString`, which should be `stringToPython` from `python_compat.h` (which actually calls `PyUnicode_FromString` when compiled for python3). This PR does that so `tf2_py` can be compiled for python3 again.

It would be nice to have a python3 platform on CI somewhere to prevent this from accidentally happening again. I'm guessing that isn't as straightforward as one might hope though.

Additionally, this PR changes some `print` statements to `print` functions in assorted python scripts for python 3 compatibility.

### DIFF
--- a/test_tf2/test/test_convert.py
+++ b/test_tf2/test/test_convert.py
@@ -34,6 +34,9 @@
 #* 
 #* Author: Eitan Marder-Eppstein
 #***********************************************************
+
+from __future__ import print_function
+
 PKG = 'test_tf2'
 import roslib; roslib.load_manifest(PKG)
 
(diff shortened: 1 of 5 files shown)


test_tf2/test/test_convert.py
```diff
@@ -34,6 +34,9 @@
 #* 
 #* Author: Eitan Marder-Eppstein
 #***********************************************************
+
+from __futute__ import print_function
```
mikaelarguedas: Looks like there is a typo: `__futute__` => `__future__`. (same below)
Nice catch for the python functions!

de-vri-es:
Woops, thanks. Fixed.

test_tf2/test/test_convert.py
de-vri-es: Woops, thanks. Fixed.

nuclearsandwich:

tfoote:
Thanks for the patch. Yeah it isn't straight forward for testing python3. We don't have an easy way to test python3 since it's not officially supported platform we don't have support for it in the ros_buildfarm targets. Starting to add the many different potential configurations starts to become combinatoric complexity for testing/support. We do get some advanced testing from users on Python3 default platforms like arch and gentoo.

de-vri-es:
> We do get some advanced testing from users on Python3 default platforms like arch and gentoo.

Yeah, I noticed this when updating `ros-kinetic-tf2-py` in our local binary Arch Linux repository at work :) I tend to notice compilation problems a lot more than runtime python errors though, since we mainly use C++ for our own projects.

PR #260, opened 2017-08-09
fix python3 import error

randoms:
I can import tf2_py now, but I cannot run the tests successfully,  lots of dependencies not work with python3 like PyKDL, rosunit.


tfoote:
Thanks for the incremental step forward. We do our best to support changes that are python3 compatible but don't test officially so there's definitely space for improvement.

PR #262, opened 2017-08-15
tf2_geometry_msgs added doTransform implementations for not stamped P…

cwecht:
…oint, Quaterion, Pose and Vector3 message types


tfoote:
Can you explain how you're using these methods? The value of the doTransform API is that it can be used automatically to make sure that data is in the right coordinate frame. Are you calling this directly?

cwecht:
Yes, I use them directly. They are useful to implement doTransform methods for other (custom) messages.

tfoote:
Ok, I guess it doesn't hurt to have them available. Could you add some basic unit tests to verify that they're working correctly and continue to work correctly in the future?

tfoote (changes requested):
Overall lgtm, it just needs some test coverage.

tfoote (approved):

Issue #263, opened 2017-08-24
interpolate function in cache.cpp outputs wrong time


vsherrod:
The interpolate function outputs the timestamp of the first transform instead of the timestamp of the time queried for.

void TimeCache::interpolate(const TransformStorage& one, const TransformStorage& two, TimePoint time, TransformStorage& output)
{
// Check for zero distance case
if( two.stamp_ == one.stamp_ )
{
output = two;
return;
}
//Calculate the ratio
tf2Scalar ratio = double((time - one.stamp_).count()) / double((two.stamp_ - one.stamp_).count());

//Interpolate translation
output.translation_.setInterpolate3(one.translation_, two.translation_, ratio);

//Interpolate rotation
output.rotation_ = slerp( one.rotation_, two.rotation_, ratio);

output.stamp_ = one.stamp_;
output.frame_id_ = one.frame_id_;
output.child_frame_id_ = one.child_frame_id_;
}

It should be:
 output.stamp_ = time;

Instead of 
output.stamp_ = one.stamp_;

tfoote:
Confirmed this is a followup to: https://discourse.ros.org/t/interpolate-function-in-cache-cpp-seems-to-output-wrong-timestamp/2489/4

We should both fix this and add a unit test to provide coverage of this output value.

tfoote:
Fixed in #265

PR #265, opened 2017-08-29
Corrected time output on interpolation function.

vsherrod:
Added unit test to check for this.

This is in response to bug #263


tfoote (changes requested):
Thanks. 

This looks like a good fix. I'd rather add the checks to the existing unit tests than add a new one. 

I think https://github.com/ros/geometry2/pull/265/files#diff-13832463cebbe4548ea9daab7935e0a7R369 and https://github.com/ros/geometry2/pull/265/files#diff-13832463cebbe4548ea9daab7935e0a7R254 would give coverage with the existing tests.

tfoote (approved):
Thanks for the update. I missed the notification of the changes.

PR #266, opened 2017-09-11
fix METH_OLDARGS is no longer supported error in python3

randoms:
can_transform_core, can_transform_full_core... cannot work with python3, [Referance](https://bugs.python.org/issue11587)


mikaelarguedas:
:+1: nice catch and finding the reference to the actual upstream ticket

Recent [Python docs](https://docs.python.org/3/c-api/structures.html#METH_KEYWORDS) have been updated to specify that `METH_KEYWORDS` **must** be combined with  `METH_VARARGS`

tfoote:
Thanks this looks good. The two tests that failed are known flakey and unreleated.

PR #267, opened 2017-09-20
Speed-up TimeCache search for large cache time.

at-wat:
The current implementation of TimeCache::findClosest is linear search,
costing pretty much for buffers with large cache time.
This commit improves search cost of findClosest by using binary search.

### DIFF
--- a/tf2/include/tf2/time_cache.h
+++ b/tf2/include/tf2/time_cache.h
@@ -34,7 +34,7 @@
 
 #include "transform_storage.h"
 
-#include <list>
+#include <deque>
 
 #include <sstream>
 
@@ -116,7 +116,7 @@ class TimeCache : public TimeCacheInterface
   
 
 private:
-  typedef std::list<TransformStorage> L_TransformStorage;
+  typedef std::deque<TransformStorage> L_TransformStorage;
   L_TransformStorage storage_;
 
   ros::Duration max_storage_time_;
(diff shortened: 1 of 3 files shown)


at-wat:
I will perform some benchmarks to confirm the performance for some use case and post here.

SirVer:
@at-wat This is great and your upstream fix is definitely the right solution. I am not sure if the ros repo is still active though.

SirVer:
@tfoote Tully, this TF makes Cartographer code real slow in some cases and this fixes it rather minimally invasive. Could you have a look?

at-wat:
I tried performance test based on tf2/test/speed_test.cpp for this change.
Following figures show the number of transforms in the buffer vs. time to lookup transform. 
- x-axis: interval (seconds) of registered transforms between Time(0) and Time(2)
- y-axis: time (seconds) to lookup transform

tf2-compare-time0
tf2-compare-time1
tf2-compare-time1 5
tf2-compare-time2

Time(0) and Time(2) means searching latest transform; binary search version has overheads.
For Time(1) and Time(1.5), lookup cost decreases for the larger number of transforms in the buffer.

The overhead seems small enough, I think.

note:
- These are done with release build.
- Updated speed-test with a time_interval option of 1.0 performs same as previous.
- raw data: https://gist.github.com/at-wat/4c19445a5bbe5224c48c51bc6d1d4a7c

tfoote:
Thanks for taking the time to benchmark. That's quite compelling. Since this is changing the members thus ABI, I'll need to look for users using this internal API that we might break if we release this into current rosdistros. If I don't find public usages of the API we can probably push this back into older distros.

PR #268, opened 2017-09-25
Fixed for cases of non-standard python install

smnogar:
This fixes an issue where for instance on mac where homebrewed python installs would result in a segmentation fault since it was using the system version of python with homebrewed extensions


tfoote (approved):
Thanks for catching that.

Issue #274, opened 2017-12-05
tf2_py compilation error with Python 3.5 


nwaniek:
The function `static PyObject *_allFramesAsDot` in `tf2_py.cpp` is incompatible with Python 3.5, as it invokes `PyString_FromString`. You can find a (trivial) patch at https://gist.github.com/rochus/aa6d274a349c6aaff692a82fa45af450

astoeckel:
Can confirm, just ran into this issue while compiling ROS on Fedora 27 in a Python 3 venv. The linked patch fixes the problem for me.

jacknlliu:
Did anyone run into the issue https://github.com/ros/geometry2/issues/259 even with this patch?

tfoote:
This same logic appears to be exactly the same as the contents of the method `stringToPython` which you're removing: https://github.com/ros/geometry2/blob/indigo-devel/tf2_py/src/python_compat.h#L17-L24 which is designed to specifically to provide this abstraction between python versions. 
```
inline PyObject *stringToPython(const char *input)
{
#if PY_MAJOR_VERSION >= 3
  return PyUnicode_FromString(input);
#else
  return PyString_FromString(input);
#endif
}
```

I'd like to see that function updated appropriately if there is an issue with it.

tfoote:
Since the proposed patch appears redundant and with the help of @allenh1, tf2_py is confirmed to be building with python 3.5 https://gist.github.com/allenh1/3e81e2a6fb28b728dc24416826e99af9 I'm going to close this until there's a reproducible issue.

PR #276, opened 2017-12-12
PyString to PyBytes fix when using Python 3

kentsommer:
PyString has been renamed to PyBytes in Python 3.x, this change remains backwards compatible with Python 2.

This fixes the issue in #274

### DIFF
--- a/tf2_py/src/tf2_py.cpp
+++ b/tf2_py/src/tf2_py.cpp
@@ -5,6 +5,10 @@
 
 #include "python_compat.h"
 
+#if PY_VERSION_HEX >= 0x03000000
+#define PyString_FromString PyBytes_FromString
+#endif
+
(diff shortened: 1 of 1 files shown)


tfoote:
The only place that I see that method being called is in the #include directly above which suggests that it won't be effected by this define?

tfoote (changes requested):
As with my comment on #274 This logic should be fixed inside the stringToPython method in the python_compat.h https://github.com/ros/geometry2/blob/indigo-devel/tf2_py/src/python_compat.h#L17-L24

This sort of just redefining functions is likely to cause maintainability issues in the long term.

tfoote:
As commented in #274 this appears to be redundant with the current code and the current code has been demonstrated to [compile under python 3.5](https://gist.github.com/allenh1/3e81e2a6fb28b728dc24416826e99af9) Closing this as I can't reproduce the issue.

PR #279, opened 2018-01-04
Replaced deprecated log macro calls

### LINKED ISSUE #271: Use of deprecated logging functions
See: https://github.com/ros/console_bridge/blob/master/include/console_bridge/console.h#L68-L90

https://github.com/ros/geometry2/search?q=logWarn

https://github.com/ros/geometry2/search?q=logError

racko:
Fixes #271

### DIFF
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -123,7 +123,7 @@ bool BufferCore::warnFrameId(const char* function_name_arg, const std::string& f
   {
     std::stringstream ss;
     ss << "Invalid argument passed to "<< function_name_arg <<" in tf2 frame_ids cannot be empty";
-    logWarn("%s",ss.str().c_str());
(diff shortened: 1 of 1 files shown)


tfoote:
For background this is being held until we fork development off of indigo

mikaelarguedas:
If you want to support recent console_bridge without/before branching out, several core packages support various console_bridge versions like this:
https://github.com/ros/class_loader/pull/52
Some packages in ROS undefined some deprecated macros as they collide so reling on the existence of the non prefixed macros is fragile. We worked around it bu using pure console_bridge calls: https://github.com/ros/class_loader/pull/66 was necessary as well

tfoote:
Merged into melodic-devel (this cannot be ported back to kinetic or earlier due to the new API only being available in console_bridge 0.3 which is not available on Jessie for Kinetic.

tfoote:
We haven't created a development fork for lunar and aren't planning to make one. Since 18.04 isn't a targeted platform I'd recommend just using the melodic branch.

PR #280, opened 2018-01-05
Fix Console Bridge 0.4 Compilation

allenh1:
* Fix call to logWarn for console bridge 0.4
* Fix call to logError for console-bridge 0.4


allenh1:
I'm not sure if the CI is going to pass or not, but this is a needed patch for using console bridge 0.4. I had to patch it when it updated on my Gentoo machine, so I figured you'd like these patches eventually.

tfoote:
Yeah, this is going to fail until we fork development off of indigo. Closing as duplicate of #279

Issue #129, opened 2015-12-09
intermittent test failure


tfoote:
The tests usually pass, but once I got this failure message: 

```
[100%] Built target _run_tests_tf2_sensor_msgs_rostest
[100%] Built target _run_tests_tf2_sensor_msgs
.[ERROR] [1449690685.406248658]: Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a seperate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.
..[ERROR] [1449690685.407016062]: Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a seperate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.
[ERROR] [1449690685.407063083]: Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a seperate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.
testtransform_listener_time_reset_test ... ok

[ROSTEST]-----------------------------------------------------------------------

[tf2_ros.rosunit-transform_listener_time_reset_test/time_backwards][FAILURE]----
/home/tfoote/work/indigo_geometry/src/geometry_experimental/tf2_ros/test/time_reset_test.cpp:92
Value of: buffer.canTransform("foo", "bar", ros::Time(101, 0))
  Actual: true
Expected: false
--------------------------------------------------------------------------------


SUMMARY
 * RESULT: FAIL
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 1

rostest log file is in /home/tfoote/.ros/log/rostest-yeti-11184.log
-- run_tests.py: verify result "/home/tfoote/work/indigo_geometry/build/test_results/tf2_ros/rostest-test_transform_listener_time_reset_test.xml"
```

It's probably a race condition in the tests. Rerunning 3 times afterwards does not reproduce the error.

tfoote:
There's also an instance here: https://github.com/ros/geometry2/pull/179#issuecomment-225039103

```
09:47:32 
09:47:32 rostest log file is in /home/buildfarm/.ros/log/rostest-162b59bda102-4132.log
09:47:32 -- run_tests.py: verify result "/tmp/catkin_workspace/test_results/test_tf2/rostest-test_static_publisher.xml"
[...]

PR #282, opened 2018-02-06
New functionality to transform PoseWithCovarianceStamped messages.

RobertBlakeAnderson:
New functions added which handle the spatial transformation of a geometry_msgs::PoseWithCovarianceStamped message. The transformation mutates the covariance matrix as well as the pose.

### DIFF
--- a/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
+++ b/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
@@ -41,6 +41,7 @@
 #include <geometry_msgs/Vector3Stamped.h>
 #include <geometry_msgs/Pose.h>
 #include <geometry_msgs/PoseStamped.h>
+#include <geometry_msgs/PoseWithCovarianceStamped.h>
 #include <kdl/frames.hpp>
 
 namespace tf2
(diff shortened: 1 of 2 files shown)


tf2_geometry_msgs/test/test_tf2_geometry_msgs.cpp
```diff
@@ -161,7 +199,8 @@ int main(int argc, char **argv){
   ros::NodeHandle n;
 
   tf_buffer = new tf2_ros::Buffer();
-
+  tf_buffer->setUsingDedicatedThread(true);
```
tfoote, line 250: Why did you have to change this?

tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
```diff
@@ -695,6 +769,118 @@ void doTransform(const geometry_msgs::PoseStamped& t_in, geometry_msgs::PoseStam
   t_out.header.frame_id = transform.header.frame_id;
 }
 
+/** \brief Transform the covariance matrix of a PoseWithCovarianceStamped message to a new frame.
+* \param t_in The covariance matrix to transform.
+* \param transform The timestamped transform to apply, as a TransformStamped message.
+* \return The transformed covariance matrix.
+*/
+geometry_msgs::PoseWithCovariance::_covariance_type transformCovariance(const geometry_msgs::PoseWithCovariance::_covariance_type& cov_in, const tf2::Transform& transform)
```
tfoote: inline missing here

tf2_geometry_msgs/test/test_tf2_geometry_msgs.cpp
```diff
@@ -74,6 +74,44 @@ TEST(TfGeometry, Frame)
...
+  v1.header.frame_id = "A";
+  v1.pose.covariance[0] = 1;
+  v1.pose.covariance[7] = 1;
+  v1.pose.covariance[13] = 1;
+  v1.pose.covariance[20] = 1;
+  v1.pose.covariance[27] = 1;
+  v1.pose.covariance[35] = 1;
+  
+  // simple api
+  const geometry_msgs::PoseWithCovarianceStamped v_simple = tf_buffer->transform(v1, "B", ros::Duration(2.0));
```
tfoote: Can you add some basic checks of the covariance values? That's a core part of this and there's no coverage of it in these tests.

tfoote (approved):
Thanks for submitting this. I had a few requests but overall it looks like a good thing to add.

tfoote (changes requested):
Sorry wrong button

tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
RobertBlakeAnderson: Will fix.

tf2_geometry_msgs/test/test_tf2_geometry_msgs.cpp
RobertBlakeAnderson: Sure thing.

tf2_geometry_msgs/test/test_tf2_geometry_msgs.cpp
RobertBlakeAnderson: When running the tests on my machine, I was getting many instances of this error:

[ERROR] [1518043130.234301067]: Do not call canTransform or lookupTransform with a timeout unless you are using another thread for populating data. Without a dedicated thread it will always timeout.  If you have a seperate thread servicing tf messages, call setUsingDedicatedThread(true) on your Buffer instance.

Adding this line eliminated the errors.

tf2_geometry_msgs/test/test_tf2_geometry_msgs.cpp
RobertBlakeAnderson: Actually, just realized the transform t used in all these tests performs no rotations. I'll need to declare a local tf_ros::Buffer and Transform for my test to do a rotation on the covariance matrix.

RobertBlakeAnderson:
Updated for requested changes. Testing the covariance transformation did reveal a bug, so good call there.

tfoote:

RobertBlakeAnderson:
Looks like the failure is due to a canTransform call timing out in the tf2_ros tests. It passes if I run the tf2_ros tests alone on my machine though.

tfoote:
Yes those are known flakey tests. See #129, I just tried to resolve them again in #283.

tf2_geometry_msgs/test/test_tf2_geometry_msgs.cpp
tfoote: It's not "correct" but if it suppresses the warnings/errors for the test console output that's probably better for readability.

tfoote (approved):
Thanks for the update and basic tests.

PR #286, opened 2018-02-19
tf2_ros::Buffer: canTransform can now deal with timeouts smaller than…

### LINKED ISSUE #281: tf2_ros/Buffer::canTransform: timout smaller than 10ms are not possible
Currently, `Buffer::canTransform`'s timeout is implemented by polling with a cycle of 0.01ms. Therefore shorter timeouts than 0.01ms are not possible.

I would like to suggest that, if `time_ou`t < 0.01ms, the polling-cycle should be adapted. 

At least, if not changed, this behavior should be documented anywhere...

cwecht:
… 10ms by using the hunderdth of the timeout for sleeping

Resolves #281 .

### DIFF
--- a/tf2_ros/src/buffer.cpp
+++ b/tf2_ros/src/buffer.cpp
@@ -128,12 +128,13 @@ Buffer::canTransform(const std::string& target_frame, const std::string& source_
 
   // poll for transform if timeout is set
   ros::Time start_time = now_fallback_to_wall();
+  const ros::Duration sleep_duration = timeout * 0.001;
   while (now_fallback_to_wall() < start_time + timeout && 
          !canTransform(target_frame, source_frame, time) &&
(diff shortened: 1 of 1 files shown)


cwecht:
hm, is it possible that `tf2_ros.rosunit-transform_listener_time_reset_test/time_backward` is a bit flaky?

tfoote:
Yeah, the test is flaky (It's the last of #129) I hoped that it was resolved in #283 I'll reopen that. 

Can you move the magic number to a const at the top and adjust it to be 100th as we talked about? Something like: `const double CAN_TRANSFORM_POLLING_SCALE = 0.01;` Polling 1000 times before the timeout seems to be expensive.

tfoote (approved):
Thanks for the iterations.

PR #297, opened 2018-06-04
Change comment style for unused doxygen

jacobperron:
This prevents the docs from overflowing into the next functions.

You can see the generated [docs for canTransform](http://docs.ros.org/melodic/api/tf2/html/classtf2_1_1BufferCore.html#a8587c2c5a900d749db3ca5aa70d18546) contain documentation from the unused lookupTwist functions. I think this change should fix it.


tfoote (approved):
Thanks for noticing and the cleanup.

PR #300, opened 2018-06-19
add PoseWithCovariance to tf2_geometry_msgs.h

fabolhak:


tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
```diff
@@ -519,6 +520,56 @@ void fromMsg(const geometry_msgs::PoseStamped& msg, tf2::Stamped<tf2::Transform>
...
+  return out;
+}
+
+/** \brief Convert a PoseWithCovariance message to its equivalent tf2 representation.
+ * This function is a specialization of the fromMsg template defined in tf2/convert.h.
+ * \param msg A PoseWithCovariance message.
+ * \param out The PoseWithCovariance converted to the equivalent tf2 type.
+ */
+inline
+void fromMsg(const geometry_msgs::PoseWithCovariance& msg, tf2::Transform& out)
```
tfoote: This is not isomorphic and thus should not be defined using this structure.

tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
```diff
@@ -519,6 +520,56 @@ void fromMsg(const geometry_msgs::PoseStamped& msg, tf2::Stamped<tf2::Transform>
...
+}
+
+/** \brief Convert as tf2 PoseWithCovariance type to its equivalent geometry_msgs representation.
+ * This function is a specialization of the toMsg template defined in tf2/convert.h.
+ * \param in An instance of the tf2::Pose specialization.
+ * \return The PoseWithCovariance converted to a geometry_msgs PoseWithCovariance message type.
+ */
+inline
+geometry_msgs::PoseWithCovariance toMsg(const tf2::Transform& in, geometry_msgs::PoseWithCovariance & out)
+{
```
tfoote: This is not isomorphic and thus should not be defined using this structure.

tfoote (changes requested):
Adding support for PoseWithCovariance is fine. But it cannot be converted into a Transform and back so that shouldn't be defined, otherwise it will compile and run, but give invalid data.

fabolhak:
Thank you for the feedback! I wasn't so sure about those two anyway. So I just removed them. Would that be ok?

I'm just a little bit confused why these conversions work for Pose and PoseWithCovariancesStamped, though.

tfoote:
> Thank you for the feedback! I wasn't so sure about those two anyway. So I just removed them. Would that be ok?

That's fine thanks.

> I'm just a little bit confused why these conversions work for Pose and PoseWithCovariancesStamped, though.

Sorry can you explain a little bit about what's confusing?

fabolhak:
I'm not really familiar with the term isomorphic. Are the conversion to and from `tf2::Transform` not isomorphic because we lose the covariance information?

But this also happens with the `PoseWithCovarianceStamped` conversions [here](https://github.com/fabolhak/geometry2/blob/27109065e4cc6a3c887b916dd0c778728e4ec56e/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h#L623) and [here](https://github.com/fabolhak/geometry2/blob/27109065e4cc6a3c887b916dd0c778728e4ec56e/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h#L638) :confused: .

tfoote:
Yes, it shouldn't loose data.

Those are both bugs that need to be resolved.

artivis:
Looks like this PR could be merged nop ?
Removing the two `PoseWithCovarianceStamped`-related erroneous conversion should happen in another PR.

ahcorde:

PR #302, opened 2018-06-26
Added WrenchStamped transformation

destogl:
Hi,

we added WranchStamped transformation we are using very often in Force-Control context. We didn't start new package as proposed in the [ROS-tutorial](https://wiki.ros.org/tf2/Tutorials/Create%20Data%20Conversion%20Package%20%28C%2B%2B%29) since I believe it should be placed with other geometry_msgs.

Should I create PRs also for kinetic?


tfoote (approved):
Thanks for the addition!

PR #305, opened 2018-06-29
Changed access to Vector to prevent memory leak

### LINKED ISSUE #304: transforming PoseStamped in Python eats RAM
We noticed that doing many transformations of a PoseStamped in Python eats up RAM. Minimum working example:

    import genpy
    import tf2_ros
    import tf2_geometry_msgs
    import geometry_msgs.msg

    def main():
        p = geometry_msgs.msg.PoseStamped()
        p.header.frame_id = 'map'

        t = geometry_msgs.msg.TransformStamped()
        t.header.frame_id = 'map'
        t.child_frame_id = 'base_link'
        t.transform.rotation.w = 1.0

        tf_buffer = tf2_ros.Buffer(debug=False)
        tf_buffer.set_transform_static(t, 'test')

        do_transform = tf_buffer.registration.get(type(p))
        transform = tf_buffer.lookup_transform('map', 'base_link', genpy.Time(0))
        while True:
            do_transform(p, transform)

    if __name__ == '__main__':
        main()

Let this run for a few seconds and you can see that RAM is eaten up. PR will follow in a bit.

mgrrx:
Fixes #304 
Not sure what exactly causes the issue but I think it's related to SIP internals:
https://github.com/orocos/orocos_kinematics_dynamics/blob/master/python_orocos_kdl/PyKDL/frames.sip#L51
Anyway, with this workaround, the issue does not appear and it should basically do the same.

### DIFF
--- a/tf2_geometry_msgs/src/tf2_geometry_msgs/tf2_geometry_msgs.py
+++ b/tf2_geometry_msgs/src/tf2_geometry_msgs/tf2_geometry_msgs.py
@@ -88,9 +88,9 @@ def do_transform_pose(pose, transform):
                                                                           pose.pose.orientation.z, pose.pose.orientation.w),
                                                 PyKDL.Vector(pose.pose.position.x, pose.pose.position.y, pose.pose.position.z))
     res = PoseStamped()
-    res.pose.position.x = f.p[0]
-    res.pose.position.y = f.p[1]
-    res.pose.position.z = f.p[2]
+    res.pose.position.x = f.p.x()
+    res.pose.position.y = f.p.y()
+    res.pose.position.z = f.p.z()
     (res.pose.orientation.x, res.pose.orientation.y, res.pose.orientation.z, res.pose.orientation.w) = f.M.GetQuaternion()
     res.header = transform.header
     return res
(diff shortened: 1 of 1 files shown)


tfoote:
I'm not sure how this is resolving the issue. I modified the test to run against Vector3Stamped and PointStamped

Which have the same logic: https://github.com/ros/geometry2/pull/305/files#diff-eadcb5e39423761228fa12275320e2e9R78
and https://github.com/ros/geometry2/pull/305/files#diff-eadcb5e39423761228fa12275320e2e9R62

But they do not experience the same memory growth.

With some testing I see the same issue.

Original
mem-graph

With the patch:
mem-graph-with-patch

mgrrx:
Hm strange, it did look like it was fixed but you are right, the issue was still there. I changed the code a little so that the translation values are directly read from the matrix. Now it does look correct for me.

About the other datastructures (VectorStamped, PointStamped): The computation is different. For PoseStampeds a kdl Frame is returned whereas in the other cases it's a vector.

mgrrx:
Can you run your tests again or tell me how you tested it so that I can reproduce the graphs?

tfoote:
I ran you test program and used https://stackoverflow.com/questions/7998302/graphing-a-processs-memory-usage/12595110#12595110 to generate the graphs. 

With you latest update it's looking a lot better. (The left edge is the left over logs from the previous run that I forgot to clear.)

mem-graph

This definitely suggests that there's a bug in the frame `p` accessor that must be creating a temporary variable that's not cleaned up /reference counted correctly.

I've created a standalone reproduction of just using the accessor and ticketed it upstream: https://github.com/orocos/orocos_kinematics_dynamics/issues/157 Since it's likely a SIP issue with no known resolution I'll go ahead and merge this as a workaround.

mgrrx:
Thank you for merging. Do you see a chance to backport it to kinetic in the near future? We are still on kinetic and the issue here is forcing us to restart a bunch of nodes over the day and in general slows down our robots when they start swapping.

tfoote:
Yes, this should get backported to kinetic too. I'll likely do a merge back of most of the changes in melodic to kinetic that are backwards compatible in the next sync cycle.

PR #307, opened 2018-07-09
preserve constness of const argument to avoid warnings

tfoote:
Replicating ros/geometry#167


tfoote:

PR #311, opened 2018-07-11
fixed nan interpolation issue

### LINKED ISSUE #310: Bug: Occasional NaN transforms
We noticed that the transforms return from tf/tf2 occasionally contain `NaN` values in the translation part. Some debugging revealed that this is caused by the interpolation logic of the time cache. The bug is hidden here:

https://github.com/ros/geometry2/blob/0761810d5bb3ffc4b6f2c832fa335f292dcce6ce/tf2/src/cache.cpp#L178-L185

The equality check compares the timestamps based on sec/nsec integers, however the `ratio` converts them to `double`. In our case the two timestamps are only a few nanoseconds apart, and the resolution after the `toSec()` conversion is not enough. Thus there is a division by zero and the interpolation returns meaningless results due to `ratio` being `NaN`.

I will open a PR with the most obvious fix: Checking that the `toSec()` difference is non-zero instead of the stamp equality check.

bluenote10:
fixes #310

### DIFF
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -176,13 +176,13 @@ uint8_t TimeCache::findClosest(TransformStorage*& one, TransformStorage*& two, r
 void TimeCache::interpolate(const TransformStorage& one, const TransformStorage& two, ros::Time time, TransformStorage& output)
 {
   // Check for zero distance case
-  if ((two.stamp_.toSec() - one.stamp_.toSec()) == 0)
+  if ((two.stamp_ - one.stamp_).toSec() == 0)
   {
     output = two;
     return;
   }
   //Calculate the ratio
-  tf2Scalar ratio = (time.toSec() - one.stamp_.toSec()) / (two.stamp_.toSec() - one.stamp_.toSec());
+  tf2Scalar ratio = (time - one.stamp_).toSec() / (two.stamp_ - one.stamp_).toSec();
 
   //Interpolate translation
   output.translation_.setInterpolate3(one.translation_, two.translation_, ratio);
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
This is working around the symptom by testing it in the same way. It would be much better to not do double comparisons of large values. 

If you instead changed the ratio to diff the times into a Duration, that would all be integer math. And then just call `Duration::toSec` on the results the problem would be avoided completely. 

And you also aren't doing floating point comparisions or extra conversions to floating point.

bluenote10:
How would you do that exactly? `Duration` does not have division and using `Duration::toSec` on the numerator and denominator separately probably doesn't fully solve the problem, because the denominator again can become zero. I could change the check to be `(two.stamp_ - one.stamp_).toSec() == 0` accordingly?

tfoote:
A floating point representation won't have a problem representing a single nanosecond and will not "become" zero. It only looses precision with large values so if you do the differencing in fixed point math. It's safe to do the division using the floating point representation of the deltas (which are small).

https://bitbashing.io/comparing-floats.html

tf2/src/cache.cpp
```diff
@@ -176,13 +176,13 @@ uint8_t TimeCache::findClosest(TransformStorage*& one, TransformStorage*& two, r
 void TimeCache::interpolate(const TransformStorage& one, const TransformStorage& two, ros::Time time, TransformStorage& output)
 {
   // Check for zero distance case
-  if( two.stamp_ == one.stamp_ )
+  if ((two.stamp_ - one.stamp_).toSec() == 0)
```
tfoote: This change is not necessary and causes noteable extra computation.

tfoote:
Thanks for the fixup, with the revert there's now a whole bunch of unrelated whitespace changes. Generally they're good, but having the unrelated changes in the diffs will make tracking across branches much harder. Please keep the diff minimal to just the code changes.

bluenote10:
Closing in favor of #313

PR #313, opened 2018-07-16
fixed nan interpoaltion issue (cleaned)

### LINKED ISSUE #310: Bug: Occasional NaN transforms

bluenote10:
fixes #310

### DIFF
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -182,7 +182,7 @@ void TimeCache::interpolate(const TransformStorage& one, const TransformStorage&
     return;
   }
   //Calculate the ratio
-  tf2Scalar ratio = (time.toSec() - one.stamp_.toSec()) / (two.stamp_.toSec() - one.stamp_.toSec());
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Great, thanks for iterating!

PR #314, opened 2018-08-13
avoid copy.

csukuangfj:


tfoote (approved):
Thanks this looks good, but since it's an API change we'll have to hold it until we fork for N Turtle.

csukuangfj:
@tfoote 
It'd better add the `const` specifier to the following functions and `ros::Time` should be passed as a `const` reference.

https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L60
https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L60
https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L74
https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L79
https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L82
https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L85

tfoote:
@csukuangfj adding const to those methods sounds good. If you could open a PR for that it would be appreciated.

Issue #315, opened 2018-08-14
default duration in TimeCache


csukuangfj:
https://github.com/ros/geometry2/blob/7ea64d25a0631fe91bd277e0d9195f18cc437151/tf2/include/tf2/time_cache.h#L99-L101
The comment `default value of 10 seconds storage ` says it saves a duration of 10 seconds, but
the value `1ULL * 1000000000LL` is 1 second, i.e., there are only 9 zeros, not 10 zeros.

tfoote:
Fixed in #317

PR #317, opened 2018-08-14
fix issue #315

csukuangfj:
fix issue #315


tfoote (approved):
Thanks good catch. The `1ULL` does look a lot like `10LL`

PR #319, opened 2018-08-14
fix issue #318

csukuangfj:
fix issue #318


tfoote (approved):
This also looks fine but like #314 it's also going to change API and in particular ABI so we'd rather not roll it out until the next release.

PR #321, opened 2018-08-19
static_transform_publisher: ignore longer than 10 arguments rather than raise error

furushchev:
After this pull request, the `static_transform_publisher` node accepts longer than 10 arguments.

### DIFF
--- a/tf2_ros/src/static_transform_broadcaster_program.cpp
+++ b/tf2_ros/src/static_transform_broadcaster_program.cpp
@@ -57,7 +57,7 @@ int main(int argc, char ** argv)
   tf2_ros::StaticTransformBroadcaster broadcaster;
   geometry_msgs::TransformStamped msg;
 
-  if(argc == 10)
+  if(argc >= 10)
   {
     msg.transform.translation.x = atof(argv[1]);
     msg.transform.translation.y = atof(argv[2]);


mikaelarguedas:
Can you please clarify in what use cases supporting more than 10 arguments is necessary ?

tfoote:
Checking that there's an expected number of arguments is done for safety. This is done to help developers avoid making mistakes by giving concise feedback when their request cannot be achieved. This is like any other input validation process such as range checking.

furushchev:
@mikaelarguedas @tfoote Thanks for your opinions!
I assume the trick to switch the publication of static topic in either TF1 or TF2 manner.

```xml
<launch>
  <arg name="use_tf_static" default="false"/>

  <!-- internal use -->
  <arg name="tf_pkg" default="tf" unless="$(arg use_tf_static)"/>
  <arg name="tf_pkg" default="tf2_ros" if="$(arg use_tf_static)"/>

  <node name="tf_pub" pkg="$(arg tf_pkg)" type="static_transform_publisher"
       args="1 0 0 0 0 0 1 origin one_meter_forward 100"/>
</launch>
```

And, I now admit myself that it is unusual case, though..

tfoote:
Yeah, since you're already doing conditional logic you should be able to find a way to conditionally add that last argument too. Since there's a relatively straightforward workaround I'm going to close this.

PR #326, opened 2018-09-13
add C++14 standard flag to packages

NeoPhix:
Have problems with compiling packages with different CMAKE_CXX_STANDARD flags in my project. Fix it by adding C++14 compile option to geometry2 package (as guaranteed in ROS melodic-devel).

### DIFF
--- a/test_tf2/CMakeLists.txt
+++ b/test_tf2/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(test_tf2)
 
+add_compile_options(-std=c++14)
+
 if(NOT CATKIN_ENABLE_TESTING)
   return()
 endif()
--- a/tf2/CMakeLists.txt
+++ b/tf2/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(tf2)
 
+add_compile_options(-std=c++14)
+
 find_package(console_bridge REQUIRED)
 find_package(catkin REQUIRED COMPONENTS geometry_msgs rostime tf2_msgs)
 find_package(Boost REQUIRED COMPONENTS signals system thread)
--- a/tf2_bullet/CMakeLists.txt
+++ b/tf2_bullet/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(tf2_bullet)
 
+add_compile_options(-std=c++14)
+
 find_package(PkgConfig REQUIRED)
 
 set(bullet_FOUND 0)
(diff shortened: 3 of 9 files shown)


tfoote (changes requested):
Can you give some more details on your platform?

All supported platforms default to using C++14 and our policy is that you should not call out the specific version of the c++ standard in every package as that would then lead to fragmentation and a huge maintenance burden going forward to keep them all in sync and compatible between the many packages. Also since it's a low water mark the branch would not be reusable in for future platforms where the default might move forward, but be backwards compatible.

NeoPhix:
Thanks for your review. I agree with policy of avoiding of fragmentation caused by different C++ standards using.

I have standard x86_64 platform with Ubuntu 16.04 (compiler: gcc 5.3) with installed ROS Kinetic Base + following packages:
    ros-kinetic-stage 
    ros-kinetic-python-qt-binding 
    ros-kinetic-resource-retriever 
    ros-kinetic-urdf 
    ros-kinetic-xacro 
    ros-kinetic-velodyne-description 
    ros-kinetic-ros-canopen 
    ros-kinetic-move-base-msgs 
    ros-kinetic-ros-control 
    ros-kinetic-control-toolbox 
    ros-kinetic-media-export

My own project contained sources of geometry, geometry2, angles and sources of some other packages (navigation stack, teb_local_planner etc.). My default build bash command was:

`catkin_make -j8 -DCMAKE_CXX_FLAGS="-march=native" -DCMAKE_C_FLAGS="-march=native" -DCMAKE_BUILD_TYPE=Release`

When I tried to upgrade version of geometry, geomtry2 and some other packages to melodic, I have got many build errors, such as:

`/forklift/ros_modules/src/std_ros/geometry2/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h:1009:55: error: ‘array’ is not a member of ‘std’
 geometry_msgs::WrenchStamped toMsg(const tf2::Stamped<std::array<tf2::Vector3, 2>>& in, geometry_msgs::WrenchStamped & out)
`
The addition to catkin_make a compilation  flag `-DCMAKE_CXX_STANDARD=11 or 14` has fixed the issue, but there is no compatibility with older C++ standards. If you suppose that this branch won't be used with packages, which support older standards by default, then these changes are really not needed.

tfoote:
We specifically call out that c++14 is a required default in the supported platforms: http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023 And this is the melodic-devel branch targeting melodic. If you're using Kinetic in general I'd recommend using the `indigo-devel` branch which is the one released into kinetic. The changes between that branch and the melodic one are mostly for system dependency changes and deprecation cycles.

Using `-DCMAKE_CXX_STANDARD=14` in your invocations is what we recommend so that all packages will get the same settings.

PR #328, opened 2018-09-14
Fix issue #327

### LINKED ISSUE #327: tf2_ros::MessageFilter non-zero time tolerance conflicts with multiple target frames
When using the the tf2_ros::MessageFilter it is possible to request that multiple target frames be transformable using `setTargetFrames()`. It is also possible to request that transforms be available at the message timestamp plus a time tolerance using `setTolerance(). However there is currently a bug in the implementation when using both features together.

The transform at the message stamp is checked here:
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L327
And the transform at the message stamp + tolerance is checked here:
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L344

Both checks happen for every frame in the `target_frames_` vector. Thus the correct number of transforms to check is `target_frames_.size() * 2` and not `target_frames_.size() + 1` as is currently being done here:
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L247
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L272

svwilliams:
Fixes #327 
* Adds a tf2_ros message_filter unittest that uses multiple target frames and a non-zero time tolerance. This unittest fails with the current code version.
* Fixes the internal `expected_success_count_` variable to be the correct value for multiple target frames and a non-zero time tolerance

As described in the issue, there is a bug when both multiple target frames and a non-zero time tolerance are used together.
The transform at the message stamp is checked here:
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L327
And the transform at the message stamp + tolerance is checked here:
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L344
Both checks happen for every frame in the `target_frames_` vector. Thus the correct number of transforms to check is `target_frames_.size() * 2` and not `target_frames_.size() + 1` as is currently being done here:
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L247
https://github.com/ros/geometry2/blob/melodic-devel/tf2_ros/include/tf2_ros/message_filter.h#L272

### DIFF
--- a/tf2_ros/CMakeLists.txt
+++ b/tf2_ros/CMakeLists.txt
@@ -114,6 +114,9 @@ add_executable(${PROJECT_NAME}_test_listener EXCLUDE_FROM_ALL test/listener_unit
 add_dependencies(${PROJECT_NAME}_test_listener ${catkin_EXPORTED_TARGETS})
 add_executable(${PROJECT_NAME}_test_time_reset EXCLUDE_FROM_ALL test/time_reset_test.cpp)
 add_dependencies(${PROJECT_NAME}_test_time_reset ${catkin_EXPORTED_TARGETS})
+add_executable(${PROJECT_NAME}_test_message_filter EXCLUDE_FROM_ALL test/message_filter_test.cpp)
+add_dependencies(${PROJECT_NAME}_test_message_filter ${catkin_EXPORTED_TARGETS})
+
 target_link_libraries(${PROJECT_NAME}_test_listener
   ${PROJECT_NAME}
   ${catkin_LIBRARIES}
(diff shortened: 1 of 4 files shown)


tfoote (approved):
Thanks for the fix and including unit test!

PR #329, opened 2018-09-18
Add check to reset buffer when rostime goes backwards

EaiPool:
When running a rosbag in a loop and the rosbag is longer than the buffer duration, you will recieve the following warning when it loops:
Warning: TF_OLD_DATA ignoring data from the past for frame {framename} at time 1.53658e+09 according to authority /play_1537197257816658254
Possible reasons are listed at http://wiki.ros.org/tf/Errors%20explained
         at line 277 in /tmp/binarydeb/ros-indigo-tf2-0.5.18/src/buffer_core.cpp
This is caused by the buffer not being reset when the rosbag loops. I have implemented a reset similar to how it is already done in the c++ code, as seen here: https://github.com/ros/geometry2/blob/f0c6ffd465619cfeb0bf89a9fd9d0ab3fb46c80e/tf2_ros/src/transform_listener.cpp#L104-L109

### DIFF
--- a/tf2_ros/src/tf2_ros/transform_listener.py
+++ b/tf2_ros/src/tf2_ros/transform_listener.py
@@ -50,6 +50,7 @@ class TransformListener():
             :param tcp_nodelay (bool) - if True, request TCP_NODELAY from publisher. Use of this option is not generally recommended in most cases as it is better to rely on timestamps in message data. Setting tcp_nodelay to True enables TCP_NODELAY for all subscribers in the same python process.
         """
         self.buffer = buffer
+        self.last_update = rospy.Time.now()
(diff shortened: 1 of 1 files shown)


tfoote (approved):
Thanks this is a good behavior to be symmetric.

EaiPool:
Great, thanks. Do you think it makes sense to also merge it into the indigo branch?

tfoote:
Yes, I'll plan to do a batch backport at some point of anything from the latest branch that is worth backporting. Many of the recent changes are good candidates for backport.

EaiPool:
Ah, great. Thanks for the info!

Issue #178, opened 2016-06-01
High check_frequency values for BufferClient break tf2


g-gemignani:
Error:

```
Traceback (most recent call last):
  File "./tf_breaker.py", line 19, in <module>
    timeout=rospy.Duration(1))
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf2_ros/buffer_client.py", line 66, in lookup_transform
    return self.__process_goal(goal)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf2_ros/buffer_client.py", line 123, in __process_goal
    return self.__process_result(self.client.get_result())
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf2_ros/buffer_client.py", line 127, in __process_result
    raise tf2.TransformException("The BufferServer returned None for result or result.error!  Something is likely wrong with the server.")
tf2.TransformException: The BufferServer returned None for result or result.error!  Something is likely wrong with the server.
```

How to reproduce it:

``` python
#!/usr/bin/env python2

import tf2_ros
import rospy

if __name__ == '__main__':
    rospy.init_node('tf_breaker')
    bc = tf2_ros.BufferClient("/tf2_buffer_server", check_frequency=1000)
    i = 0
    while True:
        time_req = rospy.Time.now()
        bc.can_transform('base_footprint',
                         'map',
                         time_req,
                         rospy.Duration())
        ts = bc.lookup_transform('base_footprint',
                                 'map',
                                 time_req,
                                 timeout=rospy.Duration(1))
        rospy.loginfo('tf lookup number %d worked!', i)
        i += 1
```

tfoote:
I can confirm that I saw this error once but cannot reproduce it again. 

```
$ python test.py 
Traceback (most recent call last):
  File "test.py", line 19, in <module>
    timeout=rospy.Duration(1))
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf2_ros/buffer_client.py", line 66, in lookup_transform
    return self.__process_goal(goal)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf2_ros/buffer_client.py", line 123, in __process_goal
    return self.__process_result(self.client.get_result())
  File "/opt/ros/indigo/lib/python2.7/dist-packages/tf2_ros/buffer_client.py", line 127, in __process_result
    raise tf2.TransformException("The BufferServer returned None for result or result.error!  Something is likely wrong with the server.")
[...]

PR #333, opened 2018-10-05
Add checks to buffer_client.py to make sure the result is available

jorgenfb:
In rare cases the result is not available when the client SUCCEEDS. The origin of the problem might be a race condition somewhere in SimpleActionClient. This fix allows the buffer client to function correctly by making sure the result is available before returning.

Related issue #178


tfoote (approved):
Thanks this should be more robust.

PR #337, opened 2018-10-27
fix(buffer-client): Use actionlib api for obtaining result and deprec…

jorgenfb:
…ate check_frequency

Use the API provided by actionlib for waiting for result. This will improve the response time and prevent problems with custom solutions (see #178). This change makes constructor parameter check_frequency obsolute and deprecates it.


tfoote:
This looks good. However could you avoid the unrelated whitespace changes in the diff?

jorgenfb:
Done! Sorry about that.

tfoote (approved):

PR #343, opened 2018-11-14
Fix python3 import error

timonegk:
In python3, relative imports are only allowed using the dot-syntax. The change is compatible to python2, too.


tfoote:
Thanks!

PR #346, opened 2018-11-22
tf2_ros: add virtual destructor to tf2_ros::BufferInterface

meyerj:
Class `tf2_ros::BufferInterface` did not define a virtual destructor, but has virtual methods. This is dangerous because it may lead to memory leaks if instances are deleted using a pointer to the base class (reference: https://stackoverflow.com/questions/1123044/when-should-your-destructor-be-virtual).

Clang and GCC with `-Wnon-virtual-dtor` or `-Weffc++` warn about this:
```
include/tf2_ros/buffer_interface.h:48:7: warning: 'tf2_ros::BufferInterface' has virtual functions but non-virtual destructor [-Wnon-virtual-dtor]                                                             
class BufferInterface                                                                                                                                                                                                                  
      ^                                                     
```

### DIFF
--- a/tf2_ros/include/tf2_ros/buffer_interface.h
+++ b/tf2_ros/include/tf2_ros/buffer_interface.h
@@ -48,6 +48,10 @@ namespace tf2_ros
 class BufferInterface
 {
 public:
+  /** \brief Virtual destructor.
+   * Every class that has virtual methods should have one.
+   */
+  virtual ~BufferInterface() {}
 
   /** \brief Get the transform between two frames by frame ID.
    * \param target_frame The frame to which data should be transformed


tfoote:
This looks right, but I believe it will break ABI so we'll need to evaulate whether an ABI change or memory leak is worse. In general since we rebuild everything right now I think the ABI change is likely less of an issue.

The CI is pending a buildfarm upgrade: https://status.ros.org

ahcorde:

meyerj:
> This PR is targeting a [EOL distribution melodic](http://wiki.ros.org/Distributions), closing this PR, feel free to reopen this against a maintained branch

Done: https://github.com/ros/geometry2/pull/560

PR #348, opened 2018-11-27
Further Python 3 compatibility

timonegk:
This fixes a few further python3 incompatibilities I ran into.


tfoote (approved):
Thanks for the patch.

PR #353, opened 2019-01-15
added parameter to advertise tf2-frames as a service, if needed

esahin90:
If you want to minimize the usage of a tf2_buffer in other nodes, this gets pretty handy to have only one node running the whole Buffer. We don't want to create a new tf_buffer to get the frame graph as a yaml, which would be advertised as a service with this parameter.


tfoote:
Thanks this looks useful.

tfoote (approved):

PR #354, opened 2019-01-17
Remove `signals` from find_package(Boost COMPONENTS ...).

de-vri-es:
tf2 is using signals2, which is not the same library. Additionally, signals2 has always been header only, and header only libraries must not be listed in `find_package(Boost COMPONENTS ...)`.

Boost 1.69 removed the old signals library entirely, so the otherwise harmless `COMPONENTS signals` actually breaks the build now.

### DIFF
--- a/tf2/CMakeLists.txt
+++ b/tf2/CMakeLists.txt
@@ -3,7 +3,7 @@ project(tf2)
 
 find_package(console_bridge REQUIRED)
 find_package(catkin REQUIRED COMPONENTS geometry_msgs rostime tf2_msgs)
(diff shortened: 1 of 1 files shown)


exoticDFT:
Are there any plans to provide a hotfix of this issue with the 0.6.5 release? I'm currently trying to build ROS melodic of Fedora 30 and was running into this issue. By default the ROS installation setup/configuration phase grabs the 0.6.5 release and still has signals in the CMakeLists.txt file. Fedora 30 and above contains boost versions that no longer contain the original Signals library (1.69+)

PR #355, opened 2019-01-19
Test for equality to None with 'is' instead of '=='

daniel-s-ingram:


tf2_ros/src/tf2_ros/buffer_interface.py
```diff
@@ -65,7 +65,7 @@ def transform(self, object_stamped, target_frame, timeout=rospy.Duration(0.0), n
         do_transform = self.registration.get(type(object_stamped))
         res = do_transform(object_stamped, self.lookup_transform(target_frame, object_stamped.header.frame_id,
                                                                  object_stamped.header.stamp, timeout))
-        if new_type == None:
+        if new_type is None:
```
tfoote:
```suggestion
        if not new_type:
```

tf2_ros/src/tf2_ros/buffer_interface.py
```diff
@@ -95,7 +95,7 @@ def transform_full(self, object_stamped, target_frame, target_time, fixed_frame,
         res = do_transform(object_stamped, self.lookup_transform_full(target_frame, target_time,
                                                                      object_stamped.header.frame_id, object_stamped.header.stamp, 
                                                                      fixed_frame, timeout))
-        if new_type == None:
+        if new_type is None:
```
tfoote:
```suggestion
        if not new_type:
```

tf2_ros/src/tf2_ros/buffer_client.py
```diff
@@ -177,9 +177,9 @@ def __process_goal(self, goal):
         return self.__process_result(self.client.get_result())
 
     def __process_result(self, result):
-        if result == None:
+        if result is None:
```
tfoote:
```suggestion
        if not result:
```

tf2_ros/src/tf2_ros/buffer_client.py
```diff
@@ -177,9 +177,9 @@ def __process_goal(self, goal):
...
             raise tf2.TransformException("The BufferServer returned None for result!  Something is likely wrong with the server.")
-        if result.error == None:
+        if result.error is None:
```
tfoote:
```suggestion
        if not result.error:
```

tfoote (changes requested):
`is None` is better than `== None` https://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html

But if we're going to clean this up we don't have a specific meanig for None so generic false check is even cleaner. https://stackoverflow.com/questions/7152441/python-if-not-val-vs-if-val-is-none

tfoote (approved):

tfoote:
Thanks for the cleanup and iteration!

PR #357, opened 2019-01-25
Move tf2::impl::toQuaternion to tf2_geometry_msgs

### LINKED ISSUE #275: Dependecy issues between tf2 and tf2_geometry_msgs
If you look closely, you can see that

tf2/include/tf2/impl/utils.h includes <tf2_geometry_msgs> and 
tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h includes <tf2>

So there is a cycle there.
I am forced to compile these packages via a different build system, and my system dies with a cyclic dependency error.
Also note how tf2 sneakily includes tf2_geometry_msgs without actually requiring the package in CMakeLists or package.xml. I fail to understand how this can build in the first place, but it surely doesn't build for me.

jspricke:
Closes: #275

This breaks the cycle between tf2 and tf2_geometry_msgs.
I didn't add deprecated attributes for functions having
tf2_geometry_msgs in the method signature, as they would need to have
the right header imported anyhow.

### DIFF
--- a/test_tf2/test/test_utils.cpp
+++ b/test_tf2/test/test_utils.cpp
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 #include <gtest/gtest.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <tf2/utils.h>
 #include <tf2_kdl/tf2_kdl.h>
-#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <ros/time.h>
 
 double epsilon = 1e-9;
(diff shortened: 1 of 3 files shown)


jspricke:
@tfoote any thoughts here?

sloretz:
Thank you for the PR!

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160). Every change comes with a risk of introducing regressions, and there isn't much time left to fix them. To make sure this PR doesn't introduce any regressions please:

* Describe how you tested this change
* Recruit at least one more person to review this PR and try it out on their system

jspricke:
Looks like this have been solved differently in https://github.com/ros2/geometry2 so if someone is interested, maybe try backporting those changes instead.

PR #360, opened 2019-02-18
Follow #337: use actionlib API in BufferClient::processGoal()

MichaelGrupp:
Follows the changes to the Python API from #337 to simplify the action client logic.

### DIFF
--- a/tf2_ros/include/tf2_ros/buffer_client.h
+++ b/tf2_ros/include/tf2_ros/buffer_client.h
@@ -56,7 +56,7 @@ namespace tf2_ros
 
       /** \brief BufferClient constructor
        * \param ns The namespace in which to look for a BufferServer
-       * \param check_frequency The frequency in Hz to check whether the BufferServer has completed a request
+       * \param check_frequency Deprecated, not used anymore
        * \param timeout_padding The amount of time to allow passed the desired timeout on the client side for communication lag
        */
       BufferClient(std::string ns, double check_frequency = 10.0, ros::Duration timeout_padding = ros::Duration(2.0));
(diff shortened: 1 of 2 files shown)


MichaelGrupp:
The second commit probably needs to be changed, depending on how you want to handle parameter deprecation.

tf2_ros/include/tf2_ros/buffer_client.h
```diff
@@ -132,7 +132,6 @@ namespace tf2_ros
       geometry_msgs::TransformStamped processGoal(const tf2_msgs::LookupTransformGoal& goal) const;
       geometry_msgs::TransformStamped processResult(const tf2_msgs::LookupTransformResult& result) const;
       mutable LookupActionClient client_;
-      double check_frequency_;
```
tfoote: If we leave this here with a note that it's unused then this change can be ABI compatible and can then be made available in active distros without need for a discussion.

tf2_ros/src/buffer_client.cpp
```diff
@@ -40,7 +40,6 @@ namespace tf2_ros
 {
   BufferClient::BufferClient(std::string ns, double check_frequency, ros::Duration timeout_padding): 
     client_(ns), 
-    check_frequency_(check_frequency),
```
tfoote: If we keep the member for compatibility we should still initialize it to avoid warnings.

tfoote (changes requested):
Thanks this is good to do as well. You're right about the 2nd commit. It's good to clean up, but stability is more valuable for now than saving the memory of a single double value.

MichaelGrupp:
Thanks for the quick review. I changed the 2nd commit to be ABI compliant.

tfoote (approved):

tfoote:
Thank you for the patch and iterating on it!

PR #361, opened 2019-02-19
export dll and exe to correct locations

kejxu:
according to catkin guidelines
- http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_libraries.html
- http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_executables.html

PR #365, opened 2019-02-19
avoid name collision with windows.h

kejxu:
`winerror.h` is implicitly included in source code and `NO_ERROR` is defined as a macro inside it. In order to use `tf2_msgs::TF2Error::NO_ERROR` correctly, `NO_ERROR` from `winerror.h` needs to be un-defined to avoid name collision


seanyen:
cc @tfoote just a friendly ping, in case this and #361 is out of your radar. Let me and @kejxu know if any feedback.

seanyen:
@tfoote Happy new year. This and #361 is ready for review and merge. Hope we can move this forward for `noetic` and `melodic`.

PR #368, opened 2019-02-27
Remove obsolete workaround for "most vexing parse" ambiguity.

MichaelGrupp:
The object initialization can't be misinterpreted as a function
declaration anymore since #353. The `+0` was a trick to avoid the most  
vexing parse problem in C++, which isn't needed anymore. Also removes
dummy classes that were probably used to experiment with said problem.

### DIFF
--- a/tf2_ros/src/buffer_server_main.cpp
+++ b/tf2_ros/src/buffer_server_main.cpp
@@ -38,20 +38,6 @@
 #include <tf2_ros/transform_listener.h>
 #include <ros/ros.h>
 
-
-class MyClass
-{
-public:
-  MyClass() {}
-  MyClass(double d) {}
-};
-
-class BlankClass
-{
-public:
-  BlankClass() {}
-};
-
 int main(int argc, char** argv)
 {
   ros::init(argc, argv, "tf_buffer");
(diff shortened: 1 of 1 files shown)


tf2_ros/src/buffer_server_main.cpp
```diff
@@ -76,13 +62,10 @@ int main(int argc, char** argv)
     node_name = "tf2_buffer_server";
   }
 
-  // WIM: this works fine:
```
tfoote, line 79: How is this change related to the above unused code? How have you verified that this change is appropriate?

MichaelGrupp:
tl, dr: the code change works because the object initialization can't be misinterpreted as a function declaration anymore since https://github.com/ros/geometry2/pull/353

---
Long explanation:

Although the code comments were not clear (and the link is dead), it looks like the original intention of Wim was to avoid the "WTF" of the [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse). I use a vscode editor with cquery extension, which directly shows the `-Wvexing-parse` warning of Clang when I try to remove the "`+0`" from the original code (g++ shows a more cryptic warning):

warning

The `+0` was just a trick to force the compiler to not treat that line as a function declaration. An extra pair of parentheses would have the same effect. And the dummy classes that I deleted look very much like dummy classes used in examples of the most vexing parse problem :smile: 

Summary: 
```c++
// warning: parentheses were disambiguated as a function declaration [-Wvexing-parse]
tf2_ros::Buffer buffer_core(ros::Duration(buffer_size));

// unambiguous
tf2_ros::Buffer buffer_core(ros::Duration(buffer_size+0));

// since #353: unambiguous because of `publish_frame_service`, but useless `+0`
tf2_ros::Buffer buffer_core(ros::Duration(buffer_size+0), publish_frame_service));

// this PR: removed `+0`, still unambiguous
tf2_ros::Buffer buffer_core(ros::Duration(buffer_size), publish_frame_service));
```

tf2_ros/src/buffer_server_main.cpp
MichaelGrupp: Yes, it works fine, see my long comment. I figured out why these things were added in the first place.

MichaelGrupp:
Updated the PR description and title.

tfoote (approved):
Perfect, thanks for the detailed explanation.

PR #369, opened 2019-03-04
Fix dangling iterator references in buffer_server.cpp

MichaelGrupp:
The iterator reference `GoalInfo& info = *it` was used after the
corresponding iterator was erased, which is undefined behavior.


MichaelGrupp:
@jspricke @uliklank

tfoote (approved):
Yup, good catch.

Issue #358, opened 2019-02-11
tf2_eigen: Should Eigen::Affine3d be converted to tf2::Transform?


peci1:
Affine transforms are a superset of ROS transforms (which are isometries). So should there really be the conversion function from Affine3d to Transform? Shouldn't the user be required to convert the Affine3d to Isometry3d in Eigen first (doing all the needed checks)? Or should the tf2 conversion function issue a warning/error when the given affine transform is not an isometry?

tfoote:
It would probably be good to deprecate that part of the API. It makes more sense since it's not an isomorphic conversion.  When first written the Affine3d was the only transform available and Isometry3d was added later and support added here recently (#335). Though we will want to target that for the next distro to not break people in the current distro.

peci1:
Do you think a build time deprecation/warning or rather a runtime warning would be suitable?

It might be nice to actually check if the given transform is an isometry or not, and issue a runtime warning when it isn't, but I can imagine that could have impact on performance. And in most cases the transform will be an isometry (because it probably comes from a tf transform).

tfoote:
A build time deprecation warning is appropriate. This can only be fixed by changing code. Spamming the user will just be annoying.

peci1:
Deprecation warnings added in #371 .

peci1:
Noetic continuation in #561.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #371, opened 2019-03-05
tf2_eigen: Deprecate methods using Affine3d.

peci1:
Based on discussion in #358 .


peci1:
The tests pass, but compiling them issues deprecation warnings because they're still testing the newly deprecated functions.

tfoote (changes requested):
This looks fine but we should switch the test cases before merging this. And for deprecating API's we'd need to target the changes into noetic and not pull them out from under people on the current release.

We haven't yet forked for noetic, but I'll add the milestone to keep track of this.

ahcorde:

peci1:
Continued in #561.

PR #374, opened 2019-03-29
add mode to subscribe only static_tf to tf2_ros

yukkysaito:
Even if we want to use only static_tf, tf is also forced to subscribe, so the computational cost becomes an issue. Added mode to subscribe only static_tf.

### DIFF
--- a/tf2_ros/include/tf2_ros/transform_listener.h
+++ b/tf2_ros/include/tf2_ros/transform_listener.h
@@ -50,16 +50,16 @@ class TransformListener
 
 public:
   /**@brief Constructor for transform listener */
-  TransformListener(tf2::BufferCore& buffer, bool spin_thread = true);
-  TransformListener(tf2::BufferCore& buffer, const ros::NodeHandle& nh, bool spin_thread = true);
+  TransformListener(tf2::BufferCore& buffer, bool spin_thread = true, bool only_static = false);
+  TransformListener(tf2::BufferCore& buffer, const ros::NodeHandle& nh, bool spin_thread = true, bool only_static = false);
 
   ~TransformListener();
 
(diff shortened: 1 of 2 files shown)


tfoote:
I don't want to break the assumptions about completeness for an arbitrary listener. In your case some or most things may work with just the static transforms. However in many cases no transforms will be available.

If you are worried about computational costs of subscribing I suggest that you take a look at using the tf2_ros buffer server and buffer client where you can have a single entity maintain the buffer and query it remotely via an action API.

yukkysaito:
Thank you!
Is there sample code somewhere?

yukkysaito:
I need to call tf at 200 Hz. Isn't the calculation cost large when using action API?

tfoote:
The tests provide relatively simple examples:

https://github.com/ros/geometry2/blob/melodic-devel/test_tf2/test/test_buffer_client.cpp
https://github.com/ros/geometry2/blob/melodic-devel/test_tf2/test/test_buffer_client.py

The server you can instanciate a standalone node or your own instance like this: 
https://github.com/ros/geometry2/blob/melodic-devel/test_tf2/test/test_buffer_server.cpp


The query itself isn't any more expensive, though there is the overhead of sending and receiving the action request and response. It might be able to support 200Hz depending on your system. But I'll point out that 200Hz seems high for a system that you're attempting to model using only static transforms.

gbiggs:
Thanks for the input, Tully. Unfortunately @yukkysaito has muddled two issues together.

Since this is something that the wider community can probably help with, I've [made a question](https://answers.ros.org/question/320231/tf2-patterns-for-high-update-rates/) explaining what we actually need to collect possible solutions.

PR #378, opened 2019-04-29
Affine->Isometry

peci1:
One more place where changing affine to isometry makes sense IMO.


tfoote (approved):
Thanks this looks good to me.

Issue #379, opened 2019-05-06
Crash when MessageFilter is destroyed while there's still a callback registered


StefanFabian:
We've noticed a crash in rviz that would occasionally happen when removing a Camera display.

Terminal output:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```

Stack trace:
```
__GI_raise(int sig) (/build/glibc-OTsEL5/glibc-2.27/sysdeps/unix/sysv/linux/raise.c:51)
__GI_abort() (/build/glibc-OTsEL5/glibc-2.27/stdlib/abort.c:79)
[Unknown/Just-In-Time compiled code] (Unknown Source:0)
std::terminate() (Unknown Source:0)
__cxa_throw (Unknown Source:0)
void boost::throw_exception<boost::lock_error>(boost::lock_error const&) (Unknown Source:0)
boost::unique_lock<boost::mutex>::lock() (Unknown Source:0)
message_filters::Signal1<sensor_msgs::Image_<std::allocator<void> > >::call(ros::MessageEvent<sensor_msgs::Image_<std::allocator<void> > const> const&) (Unknown Source:0)
tf2_ros::MessageFilter<sensor_msgs::Image_<std::allocator<void> > >::CBQueueCallback::call() (Unknown Source:0)
ros::CallbackQueue::callOneCB(ros::CallbackQueue::TLS*) (Unknown Source:0)
ros::CallbackQueue::callAvailable(ros::WallDuration) (Unknown Source:0)
ros::spinOnce() (Unknown Source:0)
rviz::VisualizationManager::onUpdate() (Unknown Source:0)
QMetaObject::activate(QObject*, int, int, void**) (Unknown Source:0)
QTimer::timeout(QTimer::QPrivateSignal) (Unknown Source:0)
QTimer::timerEvent(QTimerEvent*) (Unknown Source:0)
QObject::event(QEvent*) (Unknown Source:0)
QApplicationPrivate::notify_helper(QObject*, QEvent*) (Unknown Source:0)
QApplication::notify(QObject*, QEvent*) (Unknown Source:0)
QCoreApplication::notifyInternal2(QObject*, QEvent*) (Unknown Source:0)
```

I've traced it to the tf2_ros::MessageFilter::CBQueueCallback::call method.
[...]

PR #380, opened 2019-05-06
Fixed crash when callback is called after MessageFilter was destroyed

StefanFabian:
see #379

There's one assumption that I'm making which is that the removeByID method is a silently failing no-op if there is no callback registered for the given id.
Since it is an interface and I don't know where the actual implementation is located, I couldn't check that.

### DIFF
--- a/tf2_ros/include/tf2_ros/message_filter.h
+++ b/tf2_ros/include/tf2_ros/message_filter.h
@@ -214,6 +214,10 @@ public:
    */
   ~MessageFilter()
   {
+    if (callback_queue_)
+    {
+      callback_queue_->removeByID((uint64_t)this);
+    }
     message_connection_.disconnect();
 
     clear();


ashnap123:
Hi, thanks for your fix. I tried it and it only reduced the chance of a crash. Still getting a segfault when removing display types.

StefanFabian:
@ashnap123 That's weird since it didn't happen to me at all anymore.
How did you try to get it to crash?
I've added a camera display while running a simulation, selected a camera topic that publishes data and repeatedly enabled and disabled the display using the checkbox in the rviz displays panel.

Did you build rviz from source when trying this fix?
The fix is in a header which means any binary using this has to be recompiled.

ashnap123:
Hey, it is still crashing when you delete a camera display. Also when you delete a LaserData display. The root cause does seem to still be in the tf2_ros message_filters.

I had built rviz from source.

ashnap123:
I added this test to which uses the message filters in a similar way to the rviz plugins. With your mod this passes, without it fails.

`typedef geometry_msgs::PointStamped MESSAGE_TYPE;

void callback( const MESSAGE_TYPE::ConstPtr &msg ){

    double time = msg->header.stamp.toSec();
    ros::Duration(0.005).sleep();
    //ROS_INFO("Got message at %lf", time);

}

void createAndDestroyMessageFilters(){

    constexpr auto NUMBER_OF_MESSAGE_FILTERS_TO_CREATE = 10;
    constexpr auto ROS_SPIN_CALLS = 100;
    ros::NodeHandle rosNodeHandle;

    const std::string topicMessageWithHeader = "message_with_stamp";
    const std::string frameID = "child_frame";
    std::atomic_bool stopThreads(false);

    tf2_ros::Buffer tf2Buffer;
    tf2_ros::TransformListener tf2Listener(tf2Buffer);

    ros::Publisher publisherMessageWithStamp = rosNodeHandle.advertise<MESSAGE_TYPE>(topicMessageWithHeader, 1000);

    auto publishMessageWithStamp = [publisherMessageWithStamp, frameID, &stopThreads](){

        while( !stopThreads ) {
            MESSAGE_TYPE messgage;
            messgage.header.frame_id = frameID;
            messgage.header.stamp = ros::Time::now();
            publisherMessageWithStamp.publish(messgage);
            ros::Duration(0.001).sleep();
        }

    };

    std::thread threadPublisher( publishMessageWithStamp );

    std::shared_ptr< tf2_ros::MessageFilter< MESSAGE_TYPE > > tf2MessageFilter;
    std::shared_ptr< message_filters::Subscriber< MESSAGE_TYPE > > subscriber;

    uint counterFilterCreation = 0;

    while ( counterFilterCreation<NUMBER_OF_MESSAGE_FILTERS_TO_CREATE ){

        ROS_INFO("CREATE");

        subscriber.reset( new message_filters::Subscriber< MESSAGE_TYPE >() );
        subscriber->subscribe(rosNodeHandle, topicMessageWithHeader, 10);

        tf2MessageFilter.reset(new tf2_ros::MessageFilter<MESSAGE_TYPE>(
                *subscriber,
                tf2Buffer,
                "child_frame",
                10,
                rosNodeHandle
        ));
        tf2MessageFilter->registerCallback(boost::bind(callback, _1));

        uint counterSpin = 0;
        while( counterSpin < ROS_SPIN_CALLS ){

            ros::spinOnce();
            counterSpin++;

            ros::Duration(0.01).sleep();

        }

        ROS_INFO("DESTROY");
        tf2MessageFilter.reset();
        subscriber.reset();

        ros::spinOnce();

        counterFilterCreation++;

    }

    stopThreads = true;
    if( threadPublisher.joinable() ){
        threadPublisher.join();
    }

};`

StefanFabian:
Ah, you're right I've managed to reproduce it on my desktop.
I'm assuming removing the callback is not enough.
I'll close this pull request and make a new one with a new fix as soon as I've come up with a good solution.

StefanFabian:
@ashnap123 Okay, I've investigated some more.
I actually could **not** reproduce your issue. The actual issue was that rviz still took the global tf2_ros headers. I've quick and dirty solved this by backing up the message_filter.h header and replacing it by the modified one before building rviz.
After doing that, I could not reproduce the issue anymore.
Anyway, I had already made a different thread safe fix using a custom reference with a shared lock to allow multiple readers but only one writer (to invalidate in the destructor).
Can you verify that it works for you when replacing the global header file?

@Maintainers I have both fixes in a fork in separate branches:
If you tell me which approach you prefer, I'll make a new pull request:
Fix using a shared mutex protected pointer (a little bit worse performance but should be absolutely threadsafe): https://github.com/StefanFabian/geometry2/tree/mutex-fix
Fix using removeByID (might still have a race condition when the callback is currently preparing to execute before the destructor is finished):  https://github.com/StefanFabian/geometry2/tree/simple-fix

ashnap123:
Hi,
I think I found the problem I was having. The rviz binary lives in a privileged location. So `rviz` points to `/opt/ros/melodic/bin/rviz` so you have to start rviz with `rosrun` if you have rviz installed via package manager as well as your own build. Also a completely fresh build of your workspace is required to ensure rviz uses the latest header files. Great fix, I hope they pull it soon!

PR #382, opened 2019-05-16
tf2_tools: add monitor script

### LINKED ISSUE #256: Create a tf_monitor equivalent in tf2_tools
It can probably be just a move of the tf version into the new package with that one calling through under the hood.

timonegk:
This PR ports the tf monitor script to tf2. For the sake of unification and transparency, I would suggest to also remove the `.py` extension ot the other tf2_tools scripts.

This PR closes #256.

### DIFF
--- a/tf2_tools/CMakeLists.txt
+++ b/tf2_tools/CMakeLists.txt
@@ -2,15 +2,28 @@ cmake_minimum_required(VERSION 2.8.3)
 project(tf2_tools)
 
 find_package(catkin REQUIRED COMPONENTS tf2
-                                         tf2_msgs
-                                         tf2_ros
+                                        tf2_msgs
+                                        tf2_ros
+                                        geometry_msgs
+                                        roscpp
 )
 
 catkin_package(
    CATKIN_DEPENDS tf2
                   tf2_msgs
-                  tf2_ros)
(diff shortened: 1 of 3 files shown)


timonegk:
Ping @tfoote

timonegk:
@tfoote I know it's not great to ping a pull request, but maybe you could take a look at this one in the next time, it is almost two years old and a non-breaking change.

tfoote:
Thanks, sorry for leaving this so long without comment. I mostly haven't wanted to push this in as a new feature in the final release of the tools. Although this is a non-breaking change it increases the support burden going forward and will create parallel paths that will need to also be documented. I'd rather push ahead and get this into https://github.com/ros2/geometry2 where we can target it for future releases.

timonegk:
I understand. Since I am using the tf monitor quite often, I refactored the code in this pull request and rewrote the monitor in python. That version can be found [here](https://github.com/bit-bots/bitbots_misc/blob/391e031491b25173b913cc553df9f052f9173749/bitbots_bringup/scripts/tf_monitor.py). We are not yet using ROS 2 but will presumably switch over later this year. Then, I will probably come back with a pull request for ros2/geometry2. For now, I am closing this one.

PR #383, opened 2019-05-22
 Fixed race condition in MessageFilter using a reader-writer locked reference

StefanFabian:
This pull request fixes issue #379 using a struct that wraps the reference to the class using a reader-writer lock.
The callbacks request shared access to the mutex and can access the filter simultaneously whereas if the filter gets destroyed, the destructor requests unique access and therefore no callback can be executed while the destructor voids the reference.
Once the callback gets access to the reference, it checks if the reference was voided and simply returns Success if it was.

This pull request is an alternative to #384.
Cons:
  * More complicated and has a slight overhead due to locking

Pros:
  * Unless I've made a mistake it is definitely threadsafe and 100% fixes the issue.


rhaschke:
#379 was fixed via #402. This PR can be closed.

StefanFabian:
Thank you!

PR #384, opened 2019-05-22
Fixed race condition in MessageFilter by removing the callback

StefanFabian:
This pull request fixes issue #379 by simply removing the callbacks in the destructor.

It is an alternative to #383.
Cons:
  * Might not always fix the issue in multithreaded environments depending on the implementation/thread-safety of the CallbackQueue

Pros:
  * simple and no overhead


rhaschke (commented):
Same as #402. Can be closed here.

StefanFabian:
Thank you!
That's also my preferred solution even though it relies on the callback queue being threadsafe.

PR #400, opened 2019-08-09
Fix compile error missing ros/ros.h

sloretz:
I'm unable to build `tf2`, `tf2_bullet`, or `tf2_eigen` using `catkin_make_isolated` without this patch. The `tf2` package depends on `rostime`, so it can't include `ros/ros.h`. Doing so results in a compile failure

```
/tmp/melodic-py3/src/geometry2/tf2/test/simple_tf2_core.cpp:33:10: fatal error: ros/ros.h: No such file or directory
 #include <ros/ros.h>
          ^~~~~~~~~~~
compilation terminated.
```

This changes the include to `<ros/time.h>` from the `rostime` package, which `tf2` does depend on.

The `tf2_bullet` package doesn't need the `ros/ros.h` header at all so this removes it.

```
/tmp/melodic-py3/src/geometry2/tf2_bullet/test/test_tf2_bullet.cpp:34:10: fatal error: ros/ros.h: No such file or directory
 #include <ros/ros.h>
          ^~~~~~~~~~~
compilation terminated.
```

The `tf2_eigen` package doesn't need the header either.

```
/tmp/melodic-py3/src/geometry2/tf2_eigen/test/tf2_eigen-test.cpp:31:10: fatal error: ros/ros.h: No such file or directory
 #include <ros/ros.h>
          ^~~~~~~~~~~
compilation terminated.
```

### DIFF
--- a/tf2/test/simple_tf2_core.cpp
+++ b/tf2/test/simple_tf2_core.cpp
@@ -30,7 +30,7 @@
 #include <gtest/gtest.h>
 #include <tf2/buffer_core.h>
 #include <sys/time.h>
-#include <ros/ros.h>
+#include <ros/time.h>
 #include "tf2/LinearMath/Vector3.h"
 #include "tf2/exceptions.h"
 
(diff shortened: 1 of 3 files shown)


tfoote (approved):
Thanks, these were likely missed in the cleanup/extraction.

PR #402, opened 2019-08-13
Fix message filter

rhaschke:
This fixes a segfault of `MessageFilter` pointed out in https://github.com/ros-visualization/rviz/issues/1372 and analyzed in detail in https://github.com/ros-visualization/rviz/pull/1393.
Essentially, when destroying/clearing the filter, all pending callbacks (which store a pointer to the filter itself) should be removed. Otherwise, when called, these callbacks will access the destroyed filter pointer.
Additionally, in a destructor, only fully-qualified virtual functions may be called.

### DIFF
--- a/tf2_ros/include/tf2_ros/message_filter.h
+++ b/tf2_ros/include/tf2_ros/message_filter.h
@@ -216,7 +216,7 @@ public:
   {
     message_connection_.disconnect();
 
-    clear();
+    MessageFilter::clear();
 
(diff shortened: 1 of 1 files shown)


tfoote (approved):

rhaschke:
Thanks for merging. I just noticed that there were similar PRs, which can be closed now:
#379, #383, #384.

@tfoote, do you plan to prepare a new melodic release anytime soon?
There are a bunch of commits since the latest release. Thanks for considering.

rhaschke:
Friedly ping: https://github.com/ros/geometry2/pull/402#issuecomment-522220480

PR #404, opened 2019-08-16
Remove roslib.load_manifest

otamachan:
Hi, 
'```import roslib; roslib.load_manifest(PKG)```' seems obsolete.
I think this can be safely removed from ``melodic-devel``.
Thanks.


tfoote (approved):
Thanks yes, it's long time legacy and unused.

PR #415, opened 2019-10-09
Fix run_tests of tf2_geometry_msgs with kdl in the same workspace

Rayman:
The tests fail to compile with the following error.
```
src/geometry2/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h:47:10: fatal error: kdl/frames.hpp: No such file or directory
 #include <kdl/frames.hpp>
          ^~~~~~~~~~~~~~~~
```
The include path of orocos_kdl is found correctly, but not passed to the compilation of the tests. This commit fixes that.

Steps to reproduce:
```sh
docker run -it ros:melodic

apt update && apt install -y python-catkin-tools

mkdir -p catkin_ws/src
cd catkin_ws
catkin init
cd src

git clone https://github.com/ros/geometry2.git
git clone https://github.com/orocos/orocos_kinematics_dynamics.git


rosdep install --from-path . -iy

catkin build
catkin run_tests tf2_geometry_msgs
```

### DIFF
--- a/tf2_geometry_msgs/CMakeLists.txt
+++ b/tf2_geometry_msgs/CMakeLists.txt
@@ -15,8 +15,10 @@ catkin_package(
    CATKIN_DEPENDS  geometry_msgs tf2_ros tf2)
 
 
-include_directories(include
-                    ${catkin_INCLUDE_DIRS}
+include_directories(
+  include
+  ${orocos_kdl_INCLUDE_DIRS}
+  ${catkin_INCLUDE_DIRS}
 )
 
 link_directories(${orocos_kdl_LIBRARY_DIRS})


sloretz:
Whoops, I didn't see this before opening #419 . Would you be willing to check if #419 solves the issue on your system? It also has a fix for a similar issue in `tf2_kdl`.

Rayman:
I've tested #419 and that also fixes my problem. It's just a matter of `include_directories` vs `target_include_directories`.

tfoote:
Thanks for the PR! I found #419 first and already merged it so I'll close this as it works as well.

Issue #414, opened 2019-10-08
Transforms published with the same timestamp accumulate indefinitely in all listeners


beiju:
tf2's `TimeCache` deletes cached transforms when they are more than X seconds older than the newest transform. This means that if any node in the ROS system regularly publishes a transform with an unchanging timestamp (e.g. `ros::Time(0)`), all listeners in all nodes will increase in memory usage without bound. In practice this can cause tens of gigabytes per day.

I suggest changing the buffer so that inserting a transform with timestamp equal to an existing transform's timestamp replaces the existing transform instead of adding an additional entry. The implementation already has to search for the insertion location, so this is very minimal extra work. In the current implementation the existing transform will never be used again, so no information is lost.

If for some reason you don't want to do that, I suggest at least adding a warning the first time a transform with duplicate time stamp is received. This is a difficult issue to debug, especially considering the symptom appears in a different node than the cause.

Either way I'm happy to submit a PR if you approve.

tfoote:
Hmm, in general you're not supposed to rebroadcast data with the same timestamp. I would consider that a bug in the other code. If you want to do something like that you should be using a static transform. Note that the special capabilities of time zero are only available on query not for the publishing. Time zero represents time not initialized and nothing should actively have that timestamp. 

However we shouldn't allow the buffer to grow unbounded if someone does publish the same timestamp. The design of the tf datastructure is that it's a record of the streamed state over time and that sequential queries should be deterministic except for refinements due to delayed data. I would suggest that instead of updating the value and erroring. We drop the new value and error. 

If we allow updating I could see someone starting to do a mass republishing of content and trying to rewrite all values in the buffer since you can force a collision with any elements in the array. This is really dangerous though as if you're rewriting history and you miss a few samples you then will start interpolating over the two different time series which would be problematic. 

To have a "rewritable history" I think that we'd want to support a different datastructure similarly to how we publish tf_static in a separate datastructure a full spline series or something similar could be published for each frame and allow updating queries in the past in an holistic manner. But I am digressing from the topic at hand except to confirm that I'd like to keep the unrevisable status of the buffer. 

If you could submit a PR that checks on insert:
- If timestamp == 0 -> drop sample and provide throttled error
- If timestamp already in buffer -> drop sample and provide throttled error

pbeeson:
I independently ran into this Issue this week and submitted PR #425 (indigo/kinetic) and #426 (melodic) to fix it.

pbeeson:
[...]

PR #426, opened 2019-10-19
Don't insert a TF frame if one of the same timestamp already exists, instead just overwrite it.

pbeeson:
So, we ran into an issue with ROS on a project (still on kinetic due to legacy robot issues, but this applies to melodic too, so here's a melodic PR).

Generally, if a node is broadcasting TF data based on other incoming frames and/or other information, and something goes wrong, it will stop publishing data. But we ran into a scenario where something went wrong in a node, and it kept broadcasting the same TF data with the same timestamp, forever. When this happened, we noticed that all nodes with a tf listener started immediately consuming more and more memory.

In looking into this, it is clear that when storing an incoming TF in the tf2 listener code, that if a frame between parent and child exists with the same timestamp, that the incoming data is not ignored, nor does it overwrite the previous frame, but it gets inserted next to the existing frame with the same timestamp, so that now there are multiple transforms between the same two frames with the same timestamp.

I added some debug console outputs and a small publisher node and a node that just makes a tf listener and spins. This proved that the storage for tf listener does indeed grow infinitely so long as no newer data comes in that is more than max_duration (10 seconds) to force these to be pruned.

The fix is VERY simple, and has been tested with the stand alone programs. where previously the size of storage_ grew without bound when sending the exact same message in a loop, but now remains size 1 for that parent/child transform.

### DIFF
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -261,7 +261,10 @@ bool TimeCache::insertData(const TransformStorage& new_data)
       break;
     storage_it++;
   }
-  storage_.insert(storage_it, new_data);
+  if (storage_it->stamp_ == new_data.stamp_)
+    *storage_it = new_data;
+  else
+    storage_.insert(storage_it, new_data);
 
   pruneList();
   return true;
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -252,9 +252,9 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
     {
       if (error_str)
       {
-	std::stringstream ss;
-	ss << "TF_OLD_DATA ignoring data from the past (Possible reasons are listed at http://wiki.ros.org/tf/Errors%%20explained)";
-	*error_str = ss.str();
+        std::stringstream ss;
+        ss << "TF_OLD_DATA ignoring data from the past (Possible reasons are listed at http://wiki.ros.org/tf/Errors%%20explained)";
+        *error_str = ss.str();
       }
       return false;
     }
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -270,11 +270,11 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
   if (storage_it != storage_.end() && storage_it->stamp_ == new_data.stamp_)
   {
     if (error_str)
-      {
-	std::stringstream ss;
-	ss << "TF_REPEATED_DATA ignoring data with redundant timestamp";
-	*error_str = ss.str();
-      }
+    {
+      std::stringstream ss;
+      ss << "TF_REPEATED_DATA ignoring data with redundant timestamp";
+      *error_str = ss.str();
+    }
     return false;
   }
   else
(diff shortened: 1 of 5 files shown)


tf2/src/cache.cpp
```diff
@@ -261,7 +261,10 @@ bool TimeCache::insertData(const TransformStorage& new_data)
       break;
     storage_it++;
   }
-  storage_.insert(storage_it, new_data);
+  if (storage_it != storage_.end() && storage_it->stamp_ == new_data.stamp_)
+    *storage_it = new_data;
```
tfoote: As discussed in #414 I'd rather see the newer data dropped instead of overwriting the older data. With the current semantics `/tf` data is effectively considered immutable. And without the ability to rewrite sections of data consistently you could easily end up with data from two datasources being improperly interpolated between.

tfoote (changes requested):
Thanks for the patch, I'd like to switch it to drop instead of overwriting. But otherwise it looks good.

Adding a quick test would also be great to make sure that it keeps working.

tf2/src/cache.cpp
pbeeson: OK.  I changed the patch to return immediately when a duplicate timestamp is seen instead of overwriting data.

pbeeson: In simply changing the assignment to use the old value, one of the tests in test_tf2 was failing. Further analysis showed that it was not anchoring the forward and inverse transforms to the first frame in the list but the second one.  

So, the new code for rejecting redundant transforms actually caught a bug in the tests. :cake:

tfoote:
Great thanks for the update it looks great. I'd like to add one more element which is to add a warning that the message is being dropped. Also I'd like to check if we should be returning false from the insertData call if we don't insert. As I think that makes sense since the data wasn't inserted like the other false return case. Maybe the error can be caught one layer higher, but likely we'll have to just print a warning to console.

tf2/src/cache.cpp
tfoote: Great thanks!

pbeeson:
The issue with returning. False is that the level higher just says that the
data is old.  I didn’t want to change the function too much, but we could
return an error to be printed when false is returned, but then this level
doesn’t have all the variables that are printed for the existing message
(if I recall correctly).

pbeeson:
OK @tfoote.  I pushed a change that does accomplish both an error message AND returns false.  Because there are 2 ways false can happen now, and there are 2 different error messages, the easiest was to borrow what other functions do here and pass back an error_message to the higher level.

tfoote:
This pattern looks good for the long term with passing the, but at first pass it looks like adding the new argument is an ABI breaking change that I'd like to avoid backporting to the existing distros to avoid breaking people.

I can setup a new `noetic-devel` branch to retarget this clean implementation and then we can possibly reuse your previous implementation for the backports.

pbeeson:
That seems fair to me.  I’d prefer the previous fix (where we silently
overwrite and return true) for backporting, and the current API change for
future releases.

tfoote:
Merged in #459

flixr:
We ran into the same problem...
It would be nice to have the "original" fix (without the error string) backported to melodic.

Issue #430, opened 2019-11-28
missing symbol on convert(tf2::Quaternion, Eigen::Quaternion)


gleichdick:
I build moveit and geometry2 from source, importing moveit_commander python module throws an ImportError due to missing symbol. See ros-planning/moveit#1785. Reverting #367 fixes the issue.

With HEAD pointed to cd55e45, following snippet (extracted from [here](https://github.com/ros-planning/moveit/blob/melodic-devel/moveit_ros/planning_interface/move_group_interface/src/wrap_python_move_group.cpp)) builds fine:

```c++
#include <tf2/LinearMath/Quaternion.h>
#include <Eigen/Geometry>
#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
#include <tf2_eigen/tf2_eigen.h>
#include <iostream>

int main() {
  Eigen::Quaterniond eq;
  const tf2::Quaternion tq(0,1,0,1);

  tf2::convert(tq, eq);
  std::cout << eq.toRotationMatrix() << '\n';
}
```

With current melodic-devel (#367 applied, 1cbda1e14), the symbol is also missing:

```bash
$ g++ -I/usr/include/eigen3/ -I/ws_moveit/install/include/ -I/opt/ros/melodic/include/ test.cpp 
/tmp/ccQipwpW.o: In function `void tf2::impl::Converter<false, false>::convert<tf2::Quaternion, Eigen::Quaternion<double, 0> >(tf2::Quaternion const&, Eigen::Quaternion<double, 0>&)':
test.cpp:(.text._ZN3tf24impl9ConverterILb0ELb0EE7convertINS_10QuaternionEN5Eigen10QuaternionIdLi0EEEEEvRKT_RT0_[_ZN3tf24impl9ConverterILb0ELb0EE7convertINS_10QuaternionEN5Eigen10QuaternionIdLi0EEEEEvRKT_RT0_]+0x41): undefined reference to `void tf2::fromMsg<geometry_msgs::Quaternion_<std::allocator<void> >, Eigen::Quaternion<double, 0> >(geometry_msgs::Quaternion_<std::allocator<void> > const&, Eigen::Quaternion<double, 0>&)'
collect2: error: ld returned 1 exit status
```

The function is declared in tf2_geometry_msgs.h as `inline`, but not `template<>`. If I got it right, the declaration order is:
- `template<class A, class B> fromMsg(const A&, B&);`
- `template<class A, class B> void convert(const A&, B&) {...}`
[...]

PR #431, opened 2019-12-01
[geometry_msgs] Workaround for #430

gleichdick:
Workaround for #430. Snippet now compiles (with convert.h included last)

```c++
#include <tf2/LinearMath/Quaternion.h>
#include <Eigen/Geometry>
#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
#include <tf2_eigen/tf2_eigen.h>
#include <tf2/convert.h>
#include <iostream>

int main() {
  Eigen::Quaterniond eq;

  const tf2::Quaternion tq(0,1,0,1);
  tf2::convert(tq, eq);
  std::cout << eq.toRotationMatrix();
}
```

Note that it still depends on the include order, if convert.h is included first it does not work (compiler picks forward declaration of templated `fromMsg` which is not resolved later on).

### DIFF
--- a/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
+++ b/tf2_geometry_msgs/include/tf2_geometry_msgs/tf2_geometry_msgs.h
@@ -32,7 +32,8 @@
 #ifndef TF2_GEOMETRY_MSGS_H
 #define TF2_GEOMETRY_MSGS_H
 
-#include <tf2/convert.h>
+#include <tf2/transform_functions.h>
+#include <tf2/transform_datatypes.h>
 #include <tf2/LinearMath/Quaternion.h>
 #include <tf2/LinearMath/Transform.h>
 #include <geometry_msgs/PointStamped.h>


v4hn:
C++ includes should be independent of the include order.
Is it difficult to resolve the problem altogether?

gleichdick:
**TL;DR** Please a look at [my branch](https://github.com/gleichdick/geometry2/tree/templated_fromMsg). `toMsg` is redefined (so, client code which uses `toMsg` breaks) but `tf2::convert` should be fine. Include order does not matter anymore.


I converted all `toMsg` overloads to template specialisations with two parameters (because the return value does not count for template parameter deduction). All overloads (without `template<>`) are gone:
```c++
template<class A, class B> B& toMsg(const A&, B&);
```

For converting two non-Message types (lets say `tf2::Vector3` and `Eigen::Vector3d`), I'm writing a python script which generates type mapping (both have conversions to/from `geometry_msgs::Point` in common), like
```c++
template<>
struct commonMsgType<tf2::Vector3, Eigen::Vector3> {
    using type = geometry_msgs::Point;
};
```
This struct is then read by `tf2::convert` with some SFINAE Magic which then selects the correct `fromMsg` / `toMsg` methods.

Please note that this branch is more or less a proof of concept, so documentation and tests are missing. I'm happy to contine, but please let me know whether this solution fits your needs.

gleichdick:
replaced by #491

PR #434, opened 2020-01-16
Fixed warnings in message_filter.h

moooeeeep:
These parameter names were triggering warnings (-Wunused-parameter) with -Wall

As they appear to not be used in the function body, I commented them out...


tfoote (approved):
Thanks yes, they're unneeded in this method.

PR #435, opened 2020-01-20
do_transform_cloud now maintains the original order of the point cloud

Hi-Zed:
The faction that transforms the point cloud should keep the original order and shape of the message.
To work it requires an updated version of the "create_cloud" function proposed here:
https://github.com/ros/common_msgs/pull/157


ahcorde:
This PR is targeting an [EOL ROS distribution](http://wiki.ros.org/Distributions), Closing this PR, feel free to reopen again to a maintaned branch

PR #438, opened 2020-01-29
Add TransformListener constructors that accept TransportHints

toliver:
Add TransformListener constructors that accept TransportHints that are used by the tf topic subscriber.

This will enable the use of listeners with `ros::TransportHints().tcpNoDelay()` that will allow to receive high frequency TF messages (e.g. every 8ms) without the delay caused by Nagle's algorithm. This is relevant for some applications.

### DIFF
--- a/tf2_ros/include/tf2_ros/transform_listener.h
+++ b/tf2_ros/include/tf2_ros/transform_listener.h
@@ -50,8 +50,10 @@ class TransformListener
 
 public:
   /**@brief Constructor for transform listener */
-  TransformListener(tf2::BufferCore& buffer, bool spin_thread = true);
-  TransformListener(tf2::BufferCore& buffer, const ros::NodeHandle& nh, bool spin_thread = true);
+  TransformListener(tf2::BufferCore& buffer, bool spin_thread = true,
+      ros::TransportHints transport_hints = ros::TransportHints());
+  TransformListener(tf2::BufferCore& buffer, const ros::NodeHandle& nh, bool spin_thread = true,
+      ros::TransportHints transport_hints = ros::TransportHints());
 
   ~TransformListener();
 
@@ -73,6 +75,7 @@ private:
   ros::Subscriber message_subscriber_tf_static_;
   tf2::BufferCore& buffer_;
   bool using_dedicated_thread_;
+  ros::TransportHints transport_hints_;
   ros::Time last_update_;
  
   void dedicatedListenerThread()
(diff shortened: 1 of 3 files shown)


tf2_ros/include/tf2_ros/transform_listener.h
```diff
@@ -52,6 +52,8 @@ class TransformListener
   /**@brief Constructor for transform listener */
   TransformListener(tf2::BufferCore& buffer, bool spin_thread = true);
   TransformListener(tf2::BufferCore& buffer, const ros::NodeHandle& nh, bool spin_thread = true);
+  TransformListener(tf2::BufferCore& buffer, ros::TransportHints transport_hints, bool spin_thread = true);
```
tfoote: This is a pretty advanced feature, and I don't think that we need to support the combinatoric of both a node handle and non-nodehandle version of this API.
We might as well add it as one more optional/default argument appended to the longer signature.

tfoote (changes requested):
Well I guess if you need as low a latency as possible we should support it. 

Could you also make sure to add coverage in the unit tests for this to make sure it's at least constructed and run through the usual paces.

Note that this will need to be targeted for noetic only too.

tf2_ros/include/tf2_ros/transform_listener.h
toliver: I added an additional default argument initially, but I found that it was breaking ABI and would force me to build some other repos from source (in particular geometry/tf package).
I thought additional constructors would have less side effects and get the PR accepted more easily :)

But if that is your preferred approach I can go back to it.
In that case I would add one more optional/default argument to both constructors.

toliver:
I can add the construction with the new arguments to the tests.

So there is no way to get this merged into melodic?

If that is the case, should I then close this PR and make a new one for noetic addressing your comments?

tfoote:
With this being a lower level library that many people use throughout the system we work to avoid any ABI breaking changes which this is. If it's an urgent bugfix that would make sense, but melodic is now almost half way through it's lifespan and adding new features is generally avoided as such.

Unfortunately adding the member variable is also ABI breaking so not extending the API of the constructor isn't as big of a deal. 

You don't need to close this you should be able to retarget the PR to the other branch using the edit button at the top.

toliver:
@tfoote I have addressed your comments and retargeted to `noetic-devel`. Could you have a look at it?

It's failing to build on focal, but the error seems unrelated to the changes in this PR.

tfoote (approved):
Thanks for the cleanup it looks good.

The noetic failures are expected as not all dependencies are available in noetic.

PR #442, opened 2020-02-07
[Windows][melodic-devel] Fix install locations

seanyen:
* corrected the install locations per catkin guide.
  * http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_libraries.html
  * http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_executables.html
* generalized the install for the target of `tf2_py`.


seanyen:
@tfoote @dirk-thomas @rhaschke This is ready for review and merge. Thanks!

dirk-thomas:
~~The changes the location of installed executables - putting them in `bin` / on the `PATH` rather than placing them in `libexec where they are supposed to be invoked with `rosrun`. That sounds like a breaking / undesired change to me.~~

tfoote:
I think this is changing the install rules the `RUNTIME DESTINATION` for libraries only. Do the libraries on Windows get installed to the `RUNTIME DESTINATION` not the `LIBRARY DESTINATION`?

dirk-thomas:
> I think this is changing the install rules the `RUNTIME DESTINATION` for libraries only.

Ok, I missed that part. Sorry for the noise - ignore my previous comment.

seanyen:
> Do the libraries on Windows get installed to the `RUNTIME DESTINATION` not the `LIBRARY DESTINATION`?

@tfoote On CMake\Windows, the shared library `.DLL` is treated as a `RUNTIME` target so it gets installed to `RUNTIME DESTINATION`.

A quote from [CMake manual](https://cmake.org/cmake/help/latest/command/install.html):
> RUNTIME
Executables are treated as RUNTIME objects, except those marked with the MACOSX_BUNDLE property on macOS (see BUNDLE below.) For DLL platforms (all Windows-based systems including Cygwin), the DLL part of a shared library is treated as a RUNTIME target.

tfoote (approved):

PR #454, opened 2020-04-15
Code simplification / cleanup

rhaschke:
This is just a little bit of code cleanup I did for myself to better understand what's actually happening.
See inline comments for details.


tf2/src/buffer_core.cpp
```diff
@@ -1259,7 +1259,6 @@ void BufferCore::removeTransformableCallback(TransformableCallbackHandle handle)
   {
     boost::mutex::scoped_lock lock(transformable_requests_mutex_);
     V_TransformableRequest::iterator it = std::remove_if(transformable_requests_.begin(), transformable_requests_.end(), RemoveRequestByCallback(handle));
-    transformable_requests_.erase(it, transformable_requests_.end());
```
rhaschke, line 1262: I don't see the need for this extra `erase()`: `std::remove_if` already removes all matches.
[According to the docs](https://en.cppreference.com/w/cpp/algorithm/remove#Return_value), it might even happen that the returned iterator `it` is invalid.

tf2/src/buffer_core.cpp
```diff
@@ -1336,11 +1335,6 @@ void BufferCore::cancelTransformableRequest(TransformableRequestHandle handle)
 {
   boost::mutex::scoped_lock lock(transformable_requests_mutex_);
   V_TransformableRequest::iterator it = std::remove_if(transformable_requests_.begin(), transformable_requests_.end(), RemoveRequestByID(handle));
-
-  if (it != transformable_requests_.end())
-  {
-    transformable_requests_.erase(it, transformable_requests_.end());
```
rhaschke, line 1342: Same here.

tf2_ros/include/tf2_ros/static_transform_broadcaster.h
```diff
@@ -54,7 +54,9 @@ class StaticTransformBroadcaster{
 
   /** \brief Send a TransformStamped message
    * The stamped data structure includes frame_id, and time, and parent_id already.  */
-  void sendTransform(const geometry_msgs::TransformStamped & transform);
+  void sendTransform(const geometry_msgs::TransformStamped & transform) {
```
rhaschke, line 57: This simple API transformation might be worth an inline function.

tf2_ros/src/static_transform_broadcaster.cpp
```diff
@@ -42,36 +43,22 @@ StaticTransformBroadcaster::StaticTransformBroadcaster()
...
-  std::vector<geometry_msgs::TransformStamped> v1;
-  v1.push_back(msgtf);
-  sendTransform(v1);
-}
-
-
 void StaticTransformBroadcaster::sendTransform(const std::vector<geometry_msgs::TransformStamped> & msgtf)
 {
-  for (std::vector<geometry_msgs::TransformStamped>::const_iterator it_in = msgtf.begin(); it_in != msgtf.end(); ++it_in)
+  for (const geometry_msgs::TransformStamped& input : msgtf)
```
rhaschke, line 48: The two nested loops here kind of obscured the actual semantics of the code, namely iterating over all elements of `msgtf`, potentially finding an existing element, and updating it.
The proposed code explicitly states this semantics.

rhaschke:
Noetic fails for an unrelated reason, namely `orocos_kdl` missing.

tf2_ros/include/tf2_ros/static_transform_broadcaster.h
tfoote: Due to changing the ABI here. I'm going to retarget this for noetic

tfoote (approved):
Thanks for the cleanup. The more modern c++ definitely helps with readability. Due to the change in ABI I'm going to retarget this into noetic.

tfoote:
Merged in #461

Issue #228, opened 2017-05-02
tf2_urdf


vincentrou:
Could it be interesting to have a package tf2_urdf which can convert [urdf types](https://github.com/ros/urdfdom_headers/tree/master/urdf_model/include/urdf_model) ?

This is linked to this new package : https://github.com/ros-controls/urdf_utilities/pull/1

tfoote:
That could be an interesting thing to do. I'd suggest that it be housed in a different repository so as not to bring in a urdf dependency into the geometry2 repo.

And I'd suggest it might make sense to go side by side with your new repository urdf_utilities. However I'd suggest the catch all utilities name be dropped for a more specific name with specific scope: http://wiki.ros.org/ROS/Patterns/Conventions

tfoote:
I'm going to close this ticket here as it's not something to track in this repository. But I do encourage the creation of tf2_urdf to happen.

PR #456, opened 2020-04-19
add tf2_urdf package

standmit:
The package provides conversions between geometry_msgs types and URDF types defined [here](https://github.com/ros/urdfdom_headers/blob/master/urdf_model/include/urdf_model/pose.h).

### DIFF
--- a/geometry2/package.xml
+++ b/geometry2/package.xml
@@ -22,6 +22,7 @@
   <run_depend>tf2_ros</run_depend>
   <run_depend>tf2_sensor_msgs</run_depend>
   <run_depend>tf2_tools</run_depend>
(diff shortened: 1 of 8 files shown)


standmit:
Sorry, I made this pull request before reading [this comment](https://github.com/ros/geometry2/issues/228#issuecomment-298721892). So if it sill preferable to put this code in a separate repository, I will do it.

tfoote:
Yes, it's still preferable to do this in a separate repository. Decoupling it will make maintenance easier, and users will not notice a difference once it's released.

standmit:
Link to a separate repository: https://github.com/standmit/tf2_urdf

PR #457, opened 2020-04-20
overwrite tf2 cache entry with same timestamp

schallerr:
Currently, entries with the same timestamp are all added to `TimeCache`. This might cause the internal list to grow indefinitely, since `pruneList` will not remove any entries.

This PR fixes this problem by overwriting an existing entry with the same timestamp, instead of inserting a new entry in the list.


tfoote:
This is a duplicate of #425 and #426 Please see the longer discussion there. 

The test from this would be great to add to that too.

tfoote:
closing in favor of #459

PR #461, opened 2020-05-13
iterator cleanup

tfoote:
Using more modern C++ logic and some small cleanups from a rebase of #454 retargeting noetic. Github's retargeting really just breaks things.


kingsimba:
This PR introduced a bug in 
https://github.com/ros/geometry2/blob/c73b5939723db078c9bbe18523230ad54f859682/tf2_ros/include/tf2_ros/message_filter.h#L388

Under certain condition, it will become a dead loop.

`erase()` in `BufferCore::cancelTransformableRequest()` is necessary and should not be removed.
Because `std::remove_if()` will **only move** unwanted object to the end of the array.

Please reconsider commit https://github.com/ros/geometry2/pull/461/commits/0c3752b9009e6cb083f46261a05614c31e335f43

PR #444, opened 2020-02-08
[windows][melodic] Fix ambiguous call for tf2::convert on MSVC

seanyen:
This is another attempt to rework - `ambiguous call for tf2::convert on MSVC`. Here is the previous attempt #367.

## Problem

When building a downstream project (MoveIt), it fails to compile and reports `'tf2::fromMsg': ambiguous call to overloaded function`.

```no-highlight
wrap_python_move_group.cpp
D:\a\1\a\_ws\src\moveit\moveit_ros\planning_interface\py_bindings_tools\include\moveit/py_bindings_tools/serialize_msg.h(69): warning C4267: 'argument': conversion from 'size_t' to 'uint32_t', possible loss of data
D:\a\1\a\_ws\src\moveit\moveit_ros\planning_interface\move_group_interface\src\wrap_python_move_group.cpp(97): note: see reference to function template instantiation 'void moveit::py_bindings_tools::deserializeMsg<geometry_msgs::Pose>(const std::string &,T &)' being compiled
with
[
T=geometry_msgs::Pose
]
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(54): note: see reference to class template instantiation 'boost::arg<9>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(53): note: see reference to class template instantiation 'boost::arg<8>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(52): note: see reference to class template instantiation 'boost::arg<7>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(51): note: see reference to class template instantiation 'boost::arg<6>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(50): note: see reference to class template instantiation 'boost::arg<5>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(49): note: see reference to class template instantiation 'boost::arg<4>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(48): note: see reference to class template instantiation 'boost::arg<3>' being compiled
C:\opt\rosdeps\x64\include\boost-1_66\boost/bind/placeholders.hpp(47): note: see reference to class template instantiation 'boost::arg<2>' being compiled
[...]

PR #462, opened 2020-05-13
[windows] Fix ambiguous call for tf2::convert on MSVC

tfoote:
* rework ambiguous call on MSVC.

Forward port of #444 to noetic

### DIFF
--- a/test_tf2/CMakeLists.txt
+++ b/test_tf2/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.0.2)
 
 project(test_tf2)
 
-find_package(catkin REQUIRED COMPONENTS rosconsole roscpp rostest tf tf2 tf2_bullet tf2_ros tf2_geometry_msgs tf2_kdl tf2_msgs)
+find_package(catkin REQUIRED COMPONENTS rosconsole roscpp rostest tf tf2 tf2_bullet tf2_ros tf2_geometry_msgs tf2_kdl tf2_msgs tf2_eigen)
 find_package(Boost REQUIRED COMPONENTS thread)
 find_package(orocos_kdl REQUIRED)
 
--- a/test_tf2/package.xml
+++ b/test_tf2/package.xml
@@ -23,6 +23,7 @@
   <build_depend>tf2_geometry_msgs</build_depend>
   <build_depend>tf2_kdl</build_depend>
   <build_depend>tf2_msgs</build_depend>
+  <build_depend>tf2_eigen</build_depend>
 
   <run_depend>rosconsole</run_depend>
   <run_depend>roscpp</run_depend>
@@ -34,6 +35,7 @@
   <run_depend>tf2_geometry_msgs</run_depend>
   <run_depend>tf2_kdl</run_depend>
   <run_depend>tf2_msgs</run_depend>
+  <run_depend>tf2_eigen</run_depend>
 
   <test_depend>rosunit</test_depend>
   <test_depend>rosbash</test_depend>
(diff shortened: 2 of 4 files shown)


tfoote:
@seanyen The patch from #444 doesn't forward port to noetic successfully. Could you take a look at why?

gleichdick:
I don't want to be disrespectful, but I pointed out the problem at multiple places. Please correct me if I'm wrong.
The error message from the build is exactly the same as in #430:
```
11:56:08 test_convert.cpp:(.text._ZN3tf24impl9ConverterILb0ELb0EE7convertINS_10QuaternionEN5Eigen10QuaternionIdLi0EEEEEvRKT_RT0_[_ZN3tf24impl9ConverterILb0ELb0EE7convertINS_10QuaternionEN5Eigen10QuaternionIdLi0EEEEEvRKT_RT0_]+0x45): undefined reference to `void tf2::fromMsg<geometry_msgs::Quaternion_<std::allocator<void> >, Eigen::Quaternion<double, 0> >(geometry_msgs::Quaternion_<std::allocator<void> > const&, Eigen::Quaternion<double, 0>&)'
11:56:08 collect2: error: ld returned 1 exit status
```

IMHO, the whole conversion stuff boils down to something like (taken from [here](https://github.com/ros/geometry2/pull/444#issuecomment-584516911))
```c++
#include <iostream>
namespace msgs {
    struct A_msg { 
        A_msg(){} // for clang
    };
}

namespace datatypes {
    struct B_dt {};
}

namespace tf2 {
    template <class A, class B>
    void fromMsg(const A&, B&);

    template <class A, class B>
    void convert(const A& a, B& b) {
        fromMsg(a, b);
    }

    //template<>
    void fromMsg(const msgs::A_msg& a, datatypes::B_dt& b) {}
}

int main () {
    const msgs::A_msg a;
    datatypes::B_dt b;
    tf2::convert(a, b);
    std::cout << "Hello, world!\n";
}
```
Compilation/linking fails with GCC, Clang and MSVC (with `/permissive-` flag set). If you uncomment the `template<>` statement  __or__  if you reorder the `tf2` namespace
```c++
namespace tf2 {
    void fromMsg(const msgs::A_msg& a, datatypes::B_dt& b) {}

    template <class A, class B>
    void fromMsg(const A&, B&);

    template <class A, class B>
    void convert(const A& a, B& b) {
        fromMsg(a, b);
    }
}
```
it works fine.

When the conversion is requested, `tf2::convert()` uses the templated forward declaration for `tf2::fromMsg()` in [tf2/transform_functions.h](https://github.com/ros/geometry2/blob/3f49c20b6b318820a86667d95070dabb06da0aef/tf2/include/tf2/transform_functions.h#L99).
The point is that the corresponding function is not defined as a template specialization in [tf2_eigen.h](https://github.com/ros/geometry2/blob/3f49c20b6b318820a86667d95070dabb06da0aef/tf2_eigen/include/tf2_eigen/tf2_eigen.h#L264), but as a 'classic' overload. So, this correct overloaded function will only be chosen if it is defined before the generic forward declaration. Otherwise, these linking errors occur. The users have to take care on how they order their header `#include`s which is very error-prone.

seanyen:
@tfoote Sorry this is out of my radar. I will be checking this one soon.

sloretz:
Thank you for the PR!

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160). Every change comes with a risk of introducing regressions, and there isn't much time left to fix them. To make sure this PR doesn't introduce any regressions please:

* Describe how you tested this change
* Recruit at least one more person to review this PR and try it out on their system

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #464, opened 2020-05-15
test_tf2: remove unused tf dependency

TheDelus:
Since it is unused and all tests still work I don´t see any reason why the deprecated predecessor of tf2 should be a dependency.


tfoote (approved):
Thanks, for catching this and cleaning it up.

PR #465, opened 2020-05-30
fix clash error with tf test name

flynneva:


flynneva:
received this error during compilation from source of noetic:

```
-- tf: 1 messages, 1 services
CMake Error at catkin/cmake/test/gtest.cmake:180 (add_executable):
  add_executable cannot create target "test_transform_datatypes" because
  another target with the same name already exists.  The existing target is
  an executable created in source directory
  "/home/flynn/dev/ros/noetic/src/geometry2/tf2".  See documentation for
  policy CMP0002 for more details.
Call Stack (most recent call first):
  catkin/cmake/test/gtest.cmake:89 (_catkin_add_executable_with_google_test)
  catkin/cmake/test/gtest.cmake:37 (_catkin_add_google_test)
  geometry/tf/CMakeLists.txt:86 (catkin_add_gtest)


CMake Error at geometry/tf/CMakeLists.txt:87 (target_link_libraries):
  Attempt to add link library "tf" to target "test_transform_datatypes" which
  is not built in this directory.

  This is allowed only when policy CMP0079 is set to NEW.
```

youliangtan:
Is there a reason why this is closed? 

Fyi, I am trying to install this pkg on ubuntu22.04 from source, and it is proven that this fix works. I can reopen a similar ticket that addresses this.

tfoote:
This was closed by the original author. It's not clear how to reproduce. Test target names generally shouldn't collide with other packages.

Jeffxzj:
It always happen on every Ubuntu20.04 machine when I try to build ROS Noetic from source using catkin_make. It is weird why this PR was closed

tfoote:
@Jeffxzj can you please provide reproduction instructions?

Jeffxzj:
> @Jeffxzj can you please provide reproduction instructions?

Yes, you can reproduce this error when installing ROS Noetic from source using `catkin_make` on any Ubuntu 20.04 machine ( I have met the issue several times)
My OS environment is Ubuntu 20.04.6 LTS
Follow the instructions on ROS Wiki http://wiki.ros.org/noetic/Installation/Source
Here are the commands to install dependencies, pull source code and build using `catkin_make`
```bash
# Installing bootstrap dependencies
sudo apt-get install python3-rosdep python3-rosinstall-generator python3-vcstools python3-vcstool build-essential

# Initializing rosdep
sudo rosdep init
rosdep update

# create workspace and download source code
mkdir ~/ros_catkin_ws
cd ~/ros_catkin_ws
rosinstall_generator desktop --rosdistro noetic --deps --tar > noetic-desktop.rosinstall
mkdir ./src
vcs import --input noetic-desktop.rosinstall ./src

# Resolving Dependencies
rosdep install --from-paths ./src --ignore-packages-from-source --rosdistro noetic -y

# build 
./src/catkin/bin/catkin_make  --cmake-args -DCMAKE_BUILD_TYPE=Release
```
Then, you will get the same error in this issue. Hope these instructions will help you. 
This issue should be re-open and a PR should be merged to main.

tfoote:
Ahh, you're using `catkin_make` we've switched over to recommending `catkin_make_isolated` as can be seen in the noetic source tutorial: http://wiki.ros.org/noetic/Installation/Source The use of `catkin_make` was deprecated because of these types of potential collisions which can cause collisions outside the scope of either maintainers responsibility/awareness.

Use `./src/catkin/bin/catkin_make_isolated --install -DCMAKE_BUILD_TYPE=Release` instead and it should work. This error might be the first you hit, but it won't be the last.

Issue #468, opened 2020-06-18
BufferCore wastes time in canTransform() creating error message that is never used


peci1:
`extrapolation_error_string`
https://github.com/ros/geometry2/blob/035522926403613d6f04f148cb18cc64d2544c19/tf2/src/buffer_core.cpp#L361

is only read when `error_string` is non-null:
https://github.com/ros/geometry2/blob/035522926403613d6f04f148cb18cc64d2544c19/tf2/src/buffer_core.cpp#L461-L466

However, even when `error_string` is null, pointer to `extrapolation_error_string` is still passed to `gather()`, which means the stringstream operations are performed no matter what the value of `error_string` is.

And the string operations aren't exactly cheap (this is a flamegraph from a profiler, x axis is total time accumulated over all calls to the function):

![Snímek obrazovky pořízený 2020-06-18 15-35-40](https://user-images.githubusercontent.com/182533/85026866-847a7b00-b179-11ea-9cf4-22cbfe22e7c4.png)

peci1:
Okay, this simple case is solved by #469. But there's also a great performance loss due to the error string construction in `lookupTransform` where you can't choose to not construct the strings. I'm just thinking about an interface that would allow to tell that the user is not interested in the error string, but just in the exception type...

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #469, opened 2020-06-18
Do not waste time constructing error string if nobody is interested in it in canTransform().

peci1:
Solves #468.

Also added two more tests to speed_test to verify that this patch helps.

### DIFF
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -358,7 +358,7 @@ int BufferCore::walkToTopParent(F& f, ros::Time time, CompactFrameID target_id,
       break;
     }
 
-    CompactFrameID parent = f.gather(cache, time, &extrapolation_error_string);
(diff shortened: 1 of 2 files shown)


peci1:
Speed test before patch:

```
Info:    9 to 4
Info:    Doing 1000000 10-level 1.000000-interval tests
Info:    canTransform at Time(3) without error string took 2.520531 for an average of 0.000002521
Info:    canTransform at Time(3) with error string took 4.727922 for an average of 0.000004728
```

Speed test after patch:

```
Info:    9 to 4
Info:    Doing 1000000 10-level 1.000000-interval tests
Info:    canTransform at Time(3) without error string took 0.221355 for an average of 0.000000221
Info:    canTransform at Time(3) with error string took 4.029124 for an average of 0.000004029
```
That's tenfold speedup!

Other cases (when the transform exists) are without change.

tfoote (approved):
Thanks this is a great improvement.

peci1:
Thanks for merging. Will you pick this up to backport into Melodic, please?

PR #473, opened 2020-07-06
Use list instead of set to make build reproducible

jspricke:


jspricke:
Example: https://tests.reproducible-builds.org/debian/dbd/unstable/amd64/ros-geometry2_0.7.2-2.diffoscope.html#python--tf--geometry-msgs_-.-.---_all.deb---data.tar.xz

tfoote (approved):
Thanks. I'm surprised that set's aren't reproducible when exported. But list's are more correct from the docs anyway.

PR #475, opened 2020-07-17
Revert "do not add TF with redundant timestamp"

### LINKED ISSUE #467: New warning message for repeated transforms
I just released `robot_localization` for Noetic, and I'm noting that while the tests pass, we are getting spammed with an endless stream of this message:

```
Warning: TF_REPEATED_DATA ignoring data with redundant timestamp for frame base_link at time 1432235782.066014 according to authority unknown_publisher
         at line 278 in /tmp/binarydeb/ros-noetic-tf2-0.7.1/src/buffer_core.cpp
```

For background, this is a common scenario with `r_l`:

Cycle 1: receive measurement from sensor with time stamp `t1`. Fuse measurement. Publish state (and transform) at time `t1`.
Cycle 2: receive measurement from sensor with time stamp `t2` < `t1`. Rewind filter, insert measurement from `t2`, fuse measurement from `t2` and then measurement from `t1`. Publish corrected state at time `t1`.

From the way things look, it seems that `tf2` used to re-insert the transform, until this change:

https://github.com/ros/geometry2/commit/eefb50935bfd28223c87b3d708e916f7bcc4b4ca

That would have been helpful for the use case in question, as overwriting the first transform is what we'd want to do.

However, it was later changed:

https://github.com/ros/geometry2/commit/04625380bdff3f3e9e860fc0e85f71674ddd1587#diff-c25683741e1f7b4e50eb6c5cdcad9661R275

Now `insertData` returns `false` and prints out an error.

Either of the previous behaviors would be preferable to the new one. What are the ramifications of restoring the behavior from [this](https://github.com/ros/geometry2/commit/eefb50935bfd28223c87b3d708e916f7bcc4b4ca) commit?

Hugal31:
Allow one to send a TF with a redundant timestamp to replace it.

Closes #467

### DIFF
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -269,13 +269,7 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
   }
   if (storage_it != storage_.end() && storage_it->stamp_ == new_data.stamp_)
   {
-    if (error_str)
-    {
-      std::stringstream ss;
-      ss << "TF_REPEATED_DATA ignoring data with redundant timestamp";
-      *error_str = ss.str();
-    }
-    return false;
+    return true;
   }
   else
   {
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -267,9 +267,18 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
       break;
     storage_it++;
   }
-  storage_.insert(storage_it, new_data);
 
-  pruneList();
+  // Allow to replace TF with the same timestamp
+  if (storage_it != storage_.end() && storage_it->stamp_ == new_data.stamp_)
+  {
+    *storage_it = new_data;
+  }
+  else
+  {
+    storage_.insert(storage_it, new_data);
+    pruneList();
+  }
+
   return true;
 }
 
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -267,14 +267,7 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
       break;
     storage_it++;
   }
-  if (storage_it != storage_.end() && storage_it->stamp_ == new_data.stamp_)
-  {
-    return true;
-  }
-  else
-  {
-    storage_.insert(storage_it, new_data);
-  }
+  storage_.insert(storage_it, new_data);
 
   pruneList();
   return true;
--- a/tf2/test/cache_unittest.cpp
+++ b/tf2/test/cache_unittest.cpp
@@ -407,6 +407,30 @@ TEST(TimeCache, DuplicateEntries)
   EXPECT_TRUE(!std::isnan(stor.rotation_.w()));
 }
 
+TEST(TimeCache, AllowReplaceData)
+{
+  TimeCache cache;
+
+  TransformStorage firstTf;
+  setIdentity(firstTf);
+  firstTf.translation_.setX(0.5);
+  firstTf.frame_id_ = 3;
+  firstTf.stamp_ = ros::Time().fromNSec(1);
+
+  TransformStorage secondTf = firstTf;
+  secondTf.translation_.setX(2.0);
+
+  cache.insertData(firstTf);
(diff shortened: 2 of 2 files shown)


tf2/src/cache.cpp
```diff
@@ -267,20 +267,7 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
...
-      ss << "TF_REPEATED_DATA ignoring data with redundant timestamp";
-      *error_str = ss.str();
-    }
-    return false;
-  }
-  else
-  {
-    storage_.insert(storage_it, new_data);
-  }
+  storage_.insert(storage_it, new_data);
```
ayrton04: Would it make sense to put a comment in here stating that repeated data is considered normal so that we don't get future PRs that change the behavior back?

Also, I haven't looked at the surrounding code too deeply, but if we find a transform with the same stamp in the `storage_`, should we do this?

```
if (storage_it != storage_.end() && storage_it->stamp_ == new_data.stamp_)
{
  *storage_it = new_data;
}
else
{
  storage_.insert(storage_it, new_data);
}
```

Apologies if I'm misinterpreting what's being done here and how `storage_` is used later.

Hugal31: I'll check what pruneList do and look if I can just replace the old data.
I agree for the comment.

tf2/src/cache.cpp
```diff
@@ -267,22 +267,18 @@ bool TimeCache::insertData(const TransformStorage& new_data, std::string* error_
...
-      ss << "TF_REPEATED_DATA ignoring data with redundant timestamp";
-      *error_str = ss.str();
-    }
-    return false;
+    *storage_it = new_data;
   }
   else
   {
     storage_.insert(storage_it, new_data);
+    pruneList();
```
ayrton04, line 279: ~`pruneList` prunes based on time, it appears, so maybe we want to leave it where it was? We might want to prune in either case.~

tf2/test/cache_unittest.cpp
```diff
@@ -407,6 +407,30 @@ TEST(TimeCache, DuplicateEntries)
   EXPECT_TRUE(!std::isnan(stor.rotation_.w()));
 }
 
+TEST(TimeCache, AllowReplaceData)
```
ayrton04, line 410: :+1:

tf2/src/cache.cpp
ayrton04: Scratch that. It's using only the times in the list, not the current time.

Itamare4 (approved):

tfoote (changes requested):
Please see the discussions in #414 and #426 for why it was changed to be dropped not replaced. If you want mutable data we need to properly add support for it to make sure there's consistency. 

The previous behavior would insert both values and thus would provide discontinuous data from one side or the other. It should not be reverted.

SteveMacenski:
@tfoote in that case, we should probably just close this. For the initial ticket #467 there are suggestions to throttle the logging so its more manageable or just log the warning once on the first instance. Would you merge either of those? 

Looking at the ticket there's a total of 13 +1's from people asking for this to change. The current state of things is really nasty for users.

tfoote:
Yeah, closing this makes sense. The "nasty" state for the users is generating complaints because it's raising the warnings about unexpected behavior. We can throttle it back a little but the warning should be loud. This is a very dangerous behavior. 

The insidiousness of the issue is that it only fails to lookup sometimes and as a consequence there's multiple implementations making false assumptions about the results. And I believe that it's quite likely that the attempts to "improve" the past history are actually degrading the quality of the information for some use cases. Especially for any use cases which are sensitive to consistency or the slope since the old behavior will cause data to look like a sawtooth signal. 

For comparison we've had a similar issue with extrapolation in the past. It's "easy" to mathematically allow extrapolation so you don't have to wait for all data to come in. But then it turns out that you amplify any errors in your system, and consequently all control algorithms just get worse performance. Nothing seemed wrong and we had much better latency, but after analyzing the performance with and without extrapolation what seemed like extra power was actually worse in every use case we tested.

PR #479, opened 2020-08-27
Fix potential buffer overrun of snprintf

at-wat:


tf2/src/cache.cpp
```diff
@@ -79,7 +79,7 @@ void createExtrapolationException2(ros::Time t0, ros::Time t1, std::string* erro
     ros::Duration tdiff = t1 - t0;
     char str[141]; // Text without formatting strings has 102, each timestamp has up to 20
     snprintf(
-        str, 163,
```
at-wat, line 82: This may cause buffer overrun.

doisyg:
Can confirm this PR solved the issue https://github.com/ros/geometry2/issues/480
@tfoote this should be merged and a new deb should be released asap

tfoote (approved):
Thanks for catching that. It's much cleaner to use the sizeof.

ProfJust:
Since 2020-09-18 there is an update, that fixed this issue for me.
Just upgrade
`$ sudo apt-get dist-upgrade`

and check if the new version is installed

```
$ rosversion tf2
0.7.5
```

Issue #406, opened 2019-08-23
Two tf broadcasters in same node issue for Kinetic


brunoeducsantos:
Hi,
For python version of tf2 in Kinetic version it occurs a kind of concurrency when publishing two broadcasters in the same node:

```
  tf_broadcaster = tf2_ros.StaticTransformBroadcaster()

    static_transformStamped = TransformStamped()

    static_transformStamped.header.stamp = rospy.Time.now()
    static_transformStamped.header.frame_id = "odom"
    static_transformStamped.child_frame_id = "camera_init"

    static_transformStamped.transform.translation.x = 0
    static_transformStamped.transform.translation.y = 0
    static_transformStamped.transform.translation.z = 0

    quat = tf_conversions.transformations.quaternion_from_euler(np.radians(90),0,np.radians(90))
    static_transformStamped.transform.rotation.x = quat[0]
    static_transformStamped.transform.rotation.y = quat[1]
    static_transformStamped.transform.rotation.z = quat[2]
    static_transformStamped.transform.rotation.w = quat[3]

    tf_broadcaster.sendTransform(static_transformStamped)

 
    tf_broadcaster = tf2_ros.StaticTransformBroadcaster()
    static_transformStamped2 = TransformStamped()

    static_transformStamped2.header.stamp = rospy.Time.now()
    static_transformStamped2.header.frame_id = "camera"
    static_transformStamped2.child_frame_id = "base_link"

    static_transformStamped2.transform.translation.x = 0
    static_transformStamped2.transform.translation.y = 0
    static_transformStamped2.transform.translation.z = 0

    quat = tf_conversions.transformations.quaternion_from_euler(-np.radians(90),-np.radians(90),0)
    static_transformStamped2.transform.rotation.x = quat[0]
    static_transformStamped2.transform.rotation.y = quat[1]
    static_transformStamped2.transform.rotation.z = quat[2]
    static_transformStamped2.transform.rotation.w = quat[3]
    tf_broadcaster.sendTransform(static_transformStamped2)
```

![image](https://user-images.githubusercontent.com/3111976/63588935-9b732f80-c59f-11e9-8d99-436e1681876e.png)

Anyone has a suggestion or PR to solve this?

Regards,
Bruno

Timple:
For future reference:

This is because a `tf2_ros.StaticTransformBroadcaster()` is in fact simply a latched publisher on the `/tf_static` topic.

[...]

Issue #370, opened 2019-03-04
Add possibility to delete static TF frame.


peci1:
With tf2, there's no possibility to remove already published static TF frames. I think it'd make sense to have such possibility.

There's a relevant ARO question: https://answers.ros.org/question/311864/is-it-possible-to-delete-a-static-tfs/ .

My use-case: a sensor is removed from the robot, I regenerate the URDF, restart robot_status_publisher, but RViz and all other nodes are still showing the removed frames until they're restarted.

I'm not sure about the ideal implementation of such a thing. My idea: static_transform_broadcaster could get a `removeFrame()` or `clear()` method, remove the frame from its `net_message_` cache, and publish to `/tf_static`. The listeners could notice that they received fewer frames from this specific publisher, and seek for the ones that should be deleted from their caches. However, that'd need storing the publisher id with each transform in the cache, which is probably not a good idea. So maybe there could be a special value of `TfMessage` that'd instruct listeners to remove the frame from their caches. `NaN`s can't be this value, because `TransformListener` ignores all messages with `NaN`s and changing this behavior would break a lot of code... Infinity also doesn't seem very good... Any ideas? Maybe misusing timestamps by storing the maximum uint value?

tfoote:
Yeah, that's kind of violating the assumption of a static frame being static. The mechanism for tf being distributed doesn't allow for deleting messages based on some node stopping publishing. 

For regular transforms the timestamps won't match and this is generally how I'd recommend connecting an intermittent link with a standard frame not a static frame as the frame is not static.

If you do need to invalidate a static transform I think that disconnecting it from the tree is the appropriate method. You'll have to be careful about making sure to use different unused frame_ids and not collide.

As a potential workaround, if we added support for transforms changing from static to dynamic or vice versa then you could just publish a regular transform, that would then age out as any other transform would. 

But in general if you're doing something that has any expectation of going away it doesn't really make sense to be classified as static.

peci1:
I understand your remarks. Apparently, "static transform" can mean several things for people.

A sensor connected to a fixed location never changes its position, it can just appear/disappear. In such situation, it doesn't seem right to publish the TF at 1-100 Hz. And I think that was one of the reasons why `/tf_static` was created.

Also, the tf/tf2 system allows changing static transforms - that's why I'd expect deleting is also an option.

The workaround with disconnecting the TFs from the main tree sounds doable and I'll probably use it in our project. 

[...]

PR #484, opened 2020-10-22
Add new tf_static broadcaster class which can handle more instances per node

Timple:
Solves #406

### DIFF
--- a/tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
+++ b/tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
@@ -43,11 +43,11 @@ class MultipleStaticTransformBroadcaster(object):
     pub_tf = rospy.Publisher("/tf_static", TFMessage, queue_size=100, latch=True)
     tfs = dict()
 
-    def sendTransform(self, transform):
+    @classmethod
+    def sendTransform(cls, transform):
         if not isinstance(transform, list):
             transform = [transform]
         for tf in transform:
-            MultipleStaticTransformBroadcaster.tfs[(tf.header.frame_id, tf.child_frame_id)] = tf
+            cls.tfs[(tf.header.frame_id, tf.child_frame_id)] = tf
 
-        tfs_msg = TFMessage(MultipleStaticTransformBroadcaster.tfs.values())
-        MultipleStaticTransformBroadcaster.pub_tf.publish(tfs_msg)
+        cls.pub_tf.publish(TFMessage(cls.tfs.values()))
--- a/tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
+++ b/tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
@@ -50,4 +50,4 @@ class MultipleStaticTransformBroadcaster(object):
         for tf in transform:
             cls.tfs[(tf.header.frame_id, tf.child_frame_id)] = tf
 
-        cls.pub_tf.publish(TFMessage(cls.tfs.values()))
+        cls.pub_tf.publish(TFMessage(list(cls.tfs.values())))
--- a/tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
+++ b/tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
@@ -32,25 +32,22 @@
 
 import rospy
 from tf2_msgs.msg import TFMessage
-from geometry_msgs.msg import TransformStamped
 
 
-class StaticTransformBroadcaster(object):
+class MultipleStaticTransformBroadcaster(object):
     """
-    :class:`StaticTransformBroadcaster` is a convenient way to send static transformation on the ``"/tf_static"`` message topic.
+    :class:`MultipleStaticTransformBroadcaster`
+    keeps track of all unique transforms and resends them all to new subscribers
     """
-    already_exists = False
-
-    def __init__(self):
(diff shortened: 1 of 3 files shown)


peci1:
I guess the implemented behavior should just become the default behavior for normal static TF broadcaster, as the C++ counterpart does exactly the same... Having this duality in the interface really seems wrong to me.

peci1:
And, the other way round, the statically printed warning could get into to C++ counterpart, as the same limitation applies there, too.

Timple:
> the C++ counterpart does exactly the same

Has the same bug? Or the same fix?

peci1:
It has the behavior of aggregating all received transforms and sending the aggregate every time.

peci1:
https://github.com/ros/geometry2/blob/c73b5939723db078c9bbe18523230ad54f859682/tf2_ros/src/static_transform_broadcaster.cpp#L46-L62

Timple:
> the statically printed warning could get into to C++ counterpart, as the same limitation applies there, too.

So then the warning is not required in C++?

peci1:
No, it can still happen that you e.g. compose two nodelets, each with its own static broadcaster. The warning should thus be added to C++, too.

Or, maybe, the solution with static (global) publisher object could be ported to C++, too. But I'm not 100% sure how would that play with the plugin/nodelet loading mechanism. And adding global variables isn't a nice thing, but it could actually solve a lot of headaches and make nodelets safely composable. Do you think you could work on the C++ part or should I try to grasp it?

Timple:
Feel free to grasp it!

I won't put the effort in the c++ part since I haven't run into that yet :slightly_smiling_face:

tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
```diff
@@ -0,0 +1,53 @@
...
+    tfs = dict()
+
+    @classmethod
+    def sendTransform(cls, transform):
+        if not isinstance(transform, list):
+            transform = [transform]
+        for tf in transform:
+            cls.tfs[(tf.header.frame_id, tf.child_frame_id)] = tf
+
+        cls.pub_tf.publish(TFMessage(cls.tfs.values()))
```
peci1:
```suggestion
        cls.pub_tf.publish(TFMessage(list(cls.tfs.values())))
```

Without this change, TFMessage constructor fails on Python 3 as values() returns not directly a list, but a proxy object.

peci1:
Before this PR is merged, I've released a standalone version with the proposed solution for Python:

http://docs.ros.org/en/noetic/api/cras_py_common/html/cras.html#module-cras.static_transform_broadcaster from package http://wiki.ros.org/cras_py_common .

tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
Timple: What do you mean? We have this code in use for a year now on python3. What fails?

Timple:
Yeah, it's a bummer that we need to release custom packages with such minor fixes because they don't get merged or reviewed :slightly_frowning_face:

tf2_ros/src/tf2_ros/multiple_static_transform_broadcaster.py
peci1: Without this fix, as soon as there is a subscriber to /tf_static, the publisher starts yelling:

    Inbound TCP/IP connection failed: 'dict_values' object has no attribute 'header'

This happens on Noetic/Focal to me on several machines. On Python 3, `type({1: 2}.values())` yields type `dict_values`.

Without /tf_static subscribers, this error does not manifest.

You can try it yourself with this test node (rewrite it to use your mutli tf publisher): https://github.com/ctu-vras/ros-utils/blob/master/cras_py_common/test/publish_static_tfs.py . Run this node in one console and `rosrun tf tf_echo a c` in another one.

---

Melodic/Bionic does not have this problem becuase `type({1: 2}.values())` is directly `list`.

tfoote (commented):
Sorry for the delay. This is still pretty basic and missing some features that will provide coverage for common use cases before it gets merged into the core implementation. We'll want to put it through and API review as anything that gets added here will need to then go directly into the tick tock release cycle and as there's no upcoming releases there's no cycles available for iterating. And new features in a stable released version are generally deprioritized. 

And if we're doing aggregation there should be at least some way to also delete/remove the transforms besides unloading the python module. It might make more sense to keep a handle to each transform and reference count them to know when to start and stop the publishing any given transform.

Also it would be good to add at least a basic test so that we can know that this is working correctly and that it keeps working down the line.

Timple:
I've added this as a seperate class in the hope this would be easier to accept this change. Since that does not seem the case, should we fix the original implementation?

___

> And if we're doing aggregation there should be at least some way to also delete/remove the transforms

Not sure I agree. The standard implementation is a latched topic. Here the tf could also only be cleared by deleting the publisher . It could not be revoked. I have not seen this in practice, so why would we need that feature here?

___

> Also it would be good to add at least a basic test

Agreed! Will do

peci1:
> And if we're doing aggregation there should be at least some way to also delete/remove the transforms besides unloading the python module.

As discussed in https://github.com/ros/geometry2/issues/370, deleting static TFs is not a supported feature (and will never be in ROS 1). To work around this limitation, you can disconnect the transform from the tree (which is a use-case that should be explicitly checked in the unit tests, e.g. transform `body->wheel` changing to `nonexistent->wheel`).

PR #490, opened 2020-12-06
Mitigate flakey test in tf2_ros

at-wat:
`transform_listener_backwards_reset` fails almost 100% on my AMD EPYC
environment. Sleep by `spin_for_a_second()` seems to be too small and
slept only a millisecond.
Fix it to a second to match with the function name.
This makes the test possible to pass on EPYC environment.

### DIFF
--- a/tf2_ros/test/time_reset_test.cpp
+++ b/tf2_ros/test/time_reset_test.cpp
@@ -41,7 +41,7 @@ void spin_for_a_second()
   ros::spinOnce();
   for (uint8_t i = 0; i < 10; ++i)
   {
-    std::this_thread::sleep_for(std::chrono::microseconds(100));
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
     ros::spinOnce();
   }
 }


at-wat:
There seems be an another flaky condition (which is not fixed by this PR) at the same point of the test: http://build.ros.org/job/Npr__geometry2__ubuntu_focal_amd64/55/, http://build.ros.org/job/Npr__geometry2__ubuntu_focal_amd64/39/ and http://build.ros.org/job/Npr__geometry2__ubuntu_focal_amd64/62/.
In this case, time jump back is not detected at:
https://github.com/ros/geometry2/blob/dae26a225e6c14a5aabd25a8b08c3031e6563535/tf2_ros/src/transform_listener.cpp#L106-L111
even when the test publishes old time to `/clock` at:
https://github.com/ros/geometry2/blob/dae26a225e6c14a5aabd25a8b08c3031e6563535/tf2_ros/test/time_reset_test.cpp#L84-L85

tfoote (approved):
Slowing this down I guess makes sense the race is unfortunate. But if on a new fast machine the 10 polling cycles aren't enough it's better to wait a little bit longer.

tfoote:
> In this case, time jump back is not detected at:

Yes we're tracking time_backwards in #129 and don't have a root cause of the flakiness. Longer timeouts decrease the frequency but don't guarantee passing.

PR #492, opened 2020-12-18
tf2_ros polling interval proportional to timeout

tony2guo:
Same issue with #281, #286 only fixed it for c++, this PR fix for python.
Since polling does not require precise loop rates, we could use rospy.sleep() instead of Rate.sleep(). This could reduce some load when polling frequently.

The downside is that sometimes we don't want timeout to correlate with polling rate. Maybe adding another parameter for polling rate?

### DIFF
--- a/tf2_ros/src/tf2_ros/buffer.py
+++ b/tf2_ros/src/tf2_ros/buffer.py
@@ -33,7 +33,7 @@ import tf2_ros
 from tf2_msgs.srv import FrameGraph, FrameGraphResponse
 import rosgraph.masterapi
 
-CAN_TRANSFORM_POLLING_SCALE = 0.01
+DEFAULT_CAN_TRANSFORM_POLLING_SCALE = 0.01
 
 class Buffer(tf2.BufferCore, tf2_ros.BufferInterface):
     """
@@ -68,6 +68,8 @@ class Buffer(tf2.BufferCore, tf2_ros.BufferInterface):
                 m.lookupService('~tf2_frames')
             except (rosgraph.masterapi.Error, rosgraph.masterapi.Failure):   
                 self.frame_server = rospy.Service('~tf2_frames', FrameGraph, self.__get_frames)
+
+        self.CAN_TRANSFORM_POLLING_SCALE = DEFAULT_CAN_TRANSFORM_POLLING_SCALE
 
     def __get_frames(self, req):
        return FrameGraphResponse(self.all_frames_as_yaml()) 
@@ -118,7 +120,7 @@ class Buffer(tf2.BufferCore, tf2_ros.BufferInterface):
         """
         if timeout != rospy.Duration(0.0):
             start_time = rospy.Time.now()
-            polling_interval = timeout * CAN_TRANSFORM_POLLING_SCALE
+            polling_interval = timeout * self.CAN_TRANSFORM_POLLING_SCALE
             while (rospy.Time.now() < start_time + timeout and 
                    not self.can_transform_core(target_frame, source_frame, time)[0] and
                    (rospy.Time.now()+rospy.Duration(3.0)) >= start_time): # big jumps in time are likely bag loops, so break for them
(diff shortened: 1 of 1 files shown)


tf2_ros/src/tf2_ros/buffer.py
```diff
@@ -67,6 +67,8 @@ def __init__(self, cache_time = None, debug = True):
             except (rosgraph.masterapi.Error, rosgraph.masterapi.Failure):   
                 self.frame_server = rospy.Service('~tf2_frames', FrameGraph, self.__get_frames)
 
+        self.CAN_TRANSFORM_POLLING_SCALE = 0.01
```
tfoote: This constant should be declared at the top of the file not in the middle of the function. It also doesn't need to be a member.

tf2_ros/src/tf2_ros/buffer.py
```diff
@@ -146,11 +148,11 @@ def can_transform_full(self, target_frame, target_time, source_frame, source_tim
         """
         if timeout != rospy.Duration(0.0):
             start_time = rospy.Time.now()
-            r= rospy.Rate(20)
+            polling_interval = timeout * self.CAN_TRANSFORM_POLLING_SCALE
             while (rospy.Time.now() < start_time + timeout and 
                    not self.can_transform_full_core(target_frame, target_time, source_frame, source_time, fixed_frame)[0] and
                    (rospy.Time.now()+rospy.Duration(3.0)) >= start_time): # big jumps in time are likely bag loops, so break for them
-                r.sleep()
+                rospy.sleep(polling_interval)
```
tfoote, line 157: You imply that this is lower overhead, but you're calling rospy.sleep which is imported from rospy.timer.sleep which looks to me to be the same implementation that the rospy.timer.Rate.sleep() calls.

tfoote (changes requested):
Homogenizing the behavior with the c++ implementation makes sense. It's a slight change in that it will cause more polling for short timeouts and less for longer ones. But will make very short timeouts much more responsive. At the extremely short end the granularity of the sleep will saturate how many times it can poll.

I have a request and a question to resolve inline.

tf2_ros/src/tf2_ros/buffer.py
tony2guo: I think the overhead is contributed by calling [_remaining()](https://github.com/ros/ros_comm/blob/08fcc096f4e822b56bfb5a30b0a09ec45f0802aa/clients/rospy/src/rospy/timer.py#L103) in rospy.timer.Rate.sleep() and creating a rospy.timer.Rate object per can_transform() call. While this overhead could be ignored in most cases, calling can_transform() frequently with a small timeout would found noticeable difference.

tfoote: ok, that makes sense. And we don't need this polling to be "precise"

tf2_ros/src/tf2_ros/buffer.py
```diff
@@ -33,6 +33,8 @@
 from tf2_msgs.srv import FrameGraph, FrameGraphResponse
 import rosgraph.masterapi
 
+CAN_TRANSFORM_POLLING_SCALE = 0.01
```
tfoote: I liked having the TRANSFORM_POLLING_SCALE as a member variable. This gives the user a way to control it if they do care about it for a specific use case/module.

Please update this to use the global on for a `DEFAULT_CAN_TRANSFORM_POLLING_SCALE` which then will be used in the `__init__`  to initialize `self.CAN_TRANSFORM_POLLING_SCALE`

This will avoid magic numbers in the class, but also mean that the developer doesn't have to think about overriding a global.

tfoote (changes requested):

tfoote (approved):

tfoote:
Thanks for the fixup, retriggering ci

MatthijsBurgh (approved):

MatthijsBurgh:
@tfoote friendly ping ;)

tfoote (approved):
Thanks for iterating.

PR #502, opened 2021-03-17
Fix Python Linking on OSX

Tobias-Fischer:
In https://github.com/RoboStack/ros-noetic/issues/86 we discovered that `import tf2_py` results in a segmentation fault on OSX. We previously observed similar issues with cv_bridge: https://github.com/ros-perception/vision_opencv/pull/331 which also contains links to why the proposed solution here works (see https://github.com/pytorch/pytorch/commit/73f6715f4725a0723d8171d3131e09ac7abf0666).

### DIFF
--- a/tf2_py/CMakeLists.txt
+++ b/tf2_py/CMakeLists.txt
@@ -88,8 +88,9 @@ catkin_package(
 
 # Dynamic linking with tf worked OK, except for exception propagation, which failed in the unit test.
 # so build with the objects directly instead.
-
-link_libraries(${PYTHON_LIBRARIES})
+if(NOT APPLE)
+  link_libraries(${PYTHON_LIBRARIES})
+endif()
 add_library(tf2_py src/tf2_py.cpp)
 target_link_libraries(tf2_py ${catkin_LIBRARIES})
 add_dependencies(tf2_py ${catkin_EXPORTED_TARGETS})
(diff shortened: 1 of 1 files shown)


sonelu:
Just as a hint, might need to apply the patch also to the ROS2 stack too. I haven't been running that yet, but I assume the same would happen there.

Tobias-Fischer:
This patch was verified in [RoboStack](https://github.com/robostack/ros-noetic) where it fixes the segfault.

Tobias-Fischer:
Ping @tfoote

tf2_py/CMakeLists.txt
```diff
@@ -101,6 +102,10 @@ else()
   set_target_properties(tf2_py PROPERTIES COMPILE_FLAGS "-g -Wno-missing-field-initializers")
   set_target_properties(tf2_py PROPERTIES OUTPUT_NAME tf2 PREFIX "_" SUFFIX ".so")
 endif()
+if(APPLE)
+  set_target_properties(tf2_py PROPERTIES
+                        LINK_FLAGS "-undefined dynamic_lookup")
+endif()
```
MatthijsBurgh: Maybe put this in the else above?

Tobias-Fischer: But then it would apply to Linux too which is not what we want, is it?

MatthijsBurgh: No put the entire if in that else.

Tobias-Fischer: Ok, done.

MatthijsBurgh (approved):

MatthijsBurgh:
@tfoote friendly ping; Waiting for your approval and merging.

sloretz (approved):
Thank you for the PR!

PR #506, opened 2021-04-07
Adding date-time to frame output files

gurnarok:
Annoyed by making multiple frames, when making minute changes, only to find that I have overwritten a frame.

This PR just adds date-time tag to the output file, creating `frames_YYYY-MM-DD_HH.MM.SS.gv` and `frames_YYYY-MM-DD_HH.MM.SS.pdf` files.


tfoote (changes requested):
This is a notable behavior change to be able to release it into active distros it would have to be opt in. I'd recommend targeting this as the new version in ros2/geometry2 https://github.com/ros2/geometry2/blob/ros2/tf2_tools/tf2_tools/view_frames.py where we can land this behavior for future rosdistros.

PR #509, opened 2021-05-04
added missing package dependencies of tf2/utils

MarcelSoler:
tf2/utils.h depends on tf2_geometry_msgs https://github.com/ros/geometry2/blob/noetic-devel/tf2/include/tf2/impl/utils.h#L18
So it should be added as a package dependency.

### DIFF
--- a/tf2/CMakeLists.txt
+++ b/tf2/CMakeLists.txt
@@ -2,14 +2,14 @@ cmake_minimum_required(VERSION 3.0.2)
 project(tf2)
 
 find_package(console_bridge REQUIRED)
-find_package(catkin REQUIRED COMPONENTS geometry_msgs rostime tf2_msgs)
+find_package(catkin REQUIRED COMPONENTS geometry_msgs rostime tf2_msgs tf2_geometry_msgs)
 find_package(Boost REQUIRED COMPONENTS system thread)
 
 catkin_package(
    INCLUDE_DIRS include
    LIBRARIES tf2
    DEPENDS console_bridge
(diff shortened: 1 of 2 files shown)


mikepurvis:
@tfoote This is needed for the Nix build as well, since it's very strict about dependencies.

----

Oh ugh, I see the real issue here— it's a mutual build-depend since tf2_geometry_msgs already depends on on tf2. So an actual architectural change is needed in order to properly resolve this.

Maybe since that header doesn't actually get compiled in the tf2 package, the proper fix is just for it to be a build-export/run depend?

tfoote:
Yeah, we can't add this as it will induce a circular dependency. The build-export  dependency is probably the right dependency. Since it's a header only I think the build-export may actually be enough without the run dependency.

mikepurvis:
The `catkin_package(CATKIN_DEPENDS xxx)` macro can only accept build_depends, though, right? The real issue here is dependent packages that need to explicitly and unexpectedly pull in `tf2_geometry_msgs` to get this header— a public example I hit was `tf2_2d` from Locus:

https://github.com/locusrobotics/tf2_2d/blob/bf6d20a53ae5cbdde22c35e790917faea25189ac/CMakeLists.txt#L2-L11

This works fine when building against a base with a merged `include`, but when built in a strict environment like Nix, the explicit dep on `tf2_geometry_msgs` must be patched in.

mikepurvis:
`tf2_2d` has been resolved as of https://github.com/locusrobotics/tf2_2d/pull/4.

This ticket should probably be closed as there are other tickets like #275 and #170 which have discussed this at length already.

PR #511, opened 2021-05-11
Invert echo terminology to make tf2_tools echo.py coherent with the rest of the API

### LINKED ISSUE #510: tf2_tools echo.py use wrong terminology for source and target frame.
echo.py takes the arguments "source_frame" and "target_frame" in this order. [Further in the loop][echo.py call], it calls [`tf2_ros.Buffer.lookup_transform`][Buffer.lookup_transform] by passing `source_frame`, then `target_frame`, whereas the argument order for `lookup_transform` is target_frame then source_frame.

This is coherent with the `tf echo` script, but is not with the rest of tf and tf2 terminology, or at least the API terminology. This can (and has) lead to confusion.

I suggest to rename the two arguments so the API remains compatible. Also, providing some documentation for the command line would be nice.

[Buffer.lookup_transform]: https://github.com/ros/geometry2/blob/c73b5939723db078c9bbe18523230ad54f859682/tf2_ros/src/tf2_ros/buffer.py#L73
[echo.py call]: https://github.com/ros/geometry2/blob/c73b5939723db078c9bbe18523230ad54f859682/tf2_tools/scripts/echo.py#L170

Hugal31:
This is my proposition to improve coherence between tf2_tools echo.py and the rest of the API while not breaking backward compatibility.

If this is accepted, I will do the same on tf echo for ros/geometry.

Closes #510.

### DIFF
--- a/tf2_tools/scripts/echo.py
+++ b/tf2_tools/scripts/echo.py
@@ -167,8 +167,8 @@ class Echo():
             lookup_time = rospy.Time()
 
         try:
-            ts = self.tf_buffer.lookup_transform(self.args.source_frame,
-                                                 self.args.target_frame,
+            ts = self.tf_buffer.lookup_transform(self.args.target_frame,
+                                                 self.args.source_frame,
                                                  lookup_time)
         except tf2.LookupException as ex:
             msg = "At time {}, (current time {}) ".format(lookup_time.to_sec(), cur_time.to_sec())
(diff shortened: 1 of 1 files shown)


tfoote (changes requested):
Just swapping this around isn't going to significantly change usability and will potentially cause confusion among the existing user base. If you'd like to update the documentation to be more explicit about the object of the transformation that would be reasonable though.

tfoote:
As this can't be used in it's current form. I'm going to close it out.

PR #514, opened 2021-08-03
Longer char array for null termination needed 

lucasw:
Without the additional char messages were coming out like this:

```
Lookup would require extrapolation at time 1628007756.300918579, but only time 1628007759.293766737 is in the buffe, when looking up transform from frame ...
                                                                                                              ^^^^^
```

`buffe`  is supposed to be `buffer`, which is fixed by extending the array the error string was put into.  It looks like the other lines of similar code for 'into the past' and 'into the future' are doing it properly already.

Should be good until the year 2286 now... but maybe this could be avoided with a macro, less brittle if someone edits the error text of changes the digits of precision?

### DIFF
--- a/tf2/src/cache.cpp
+++ b/tf2/src/cache.cpp
@@ -66,7 +66,7 @@ void createExtrapolationException1(ros::Time t0, ros::Time t1, std::string* erro
 {
   if (error_str)
   {
-    char str[116]; // Text without formatting strings has 76, each timestamp has up to 20
+    char str[117]; // Text without formatting strings has 76, each timestamp has up to 20
     snprintf(str, sizeof(str), "Lookup would require extrapolation at time %.09f, but only time %.09f is in the buffer", t0.toSec(), t1.toSec());
     *error_str = str;
   }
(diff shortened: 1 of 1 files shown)


tf2/src/cache.cpp
```diff
@@ -66,7 +66,7 @@ void createExtrapolationException1(ros::Time t0, ros::Time t1, std::string* erro
 {
   if (error_str)
   {
-    char str[116]; // Text without formatting strings has 76, each timestamp has up to 20
+    char str[117]; // Text without formatting strings has 76, each timestamp has up to 20
```
tfoote:
```suggestion
    char str[117]; // Text without formatting strings has 77, each timestamp has up to 20
```
Fix doc too

tfoote (approved):
Good catch.

peci1 (approved):

ahcorde:
@tfoote @peci1 This was approved but nerve merged. Can we merge this PR ?

peci1:
My approval is still valid ;)

ahcorde (approved):

PR #516, opened 2021-08-19
throttle TF_REPEATED_DATA to 10 seconds

c-andy-martin:
The TF_REPEATED_DATA error, while important, can be overly verbose,
sometimes flooding the terminal scrollback buffer such that no other
output can be found. Throttle the error to 10 seconds. Implement the
throttling directly since console bridge does not implement throttling.
Because ros::Time may not yet be initialized when calling insertData,
use the stamp from the message as "now".

For more information including the justification for the warning and
the justification for a 10 second throttle, read the discussion on
issue #467:

https://github.com/ros/geometry2/issues/467

### DIFF
--- a/tf2/include/tf2/time_cache.h
+++ b/tf2/include/tf2/time_cache.h
@@ -118,7 +118,7 @@ private:
   L_TransformStorage storage_;
 
   ros::Duration max_storage_time_;
-
+  ros::Time last_repeated_warn_time_;
 
   /// A helper function for getData
   //Assumes storage is already locked for it
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -275,7 +275,10 @@ bool BufferCore::setTransform(const geometry_msgs::TransformStamped& transform_i
     }
     else
     {
-      CONSOLE_BRIDGE_logWarn((error_string+" for frame %s at time %lf according to authority %s").c_str(), stripped.child_frame_id.c_str(), stripped.header.stamp.toSec(), authority.c_str());
+      if (error_string.size())
+      {
+        CONSOLE_BRIDGE_logWarn((error_string+" for frame %s at time %lf according to authority %s").c_str(), stripped.child_frame_id.c_str(), stripped.header.stamp.toSec(), authority.c_str());
+      }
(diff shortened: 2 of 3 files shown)


icefoxen:
Please Eris yes.  It would be nice to be able to upgrade a system from Melodic to Noetic and see the actual error messages, instead of literally spamming the logfiles into lagging out because someone decided that "works mostly correctly" is equivalent to "fatal error" except without the politeness of actually stopping the erroneous process.

klaxalk:
Yes, please, I totally second this. The warning is overly verbose, especially, when simulating.

tf2/include/tf2/time_cache.h
```diff
@@ -118,7 +118,7 @@ class TimeCache : public TimeCacheInterface
   L_TransformStorage storage_;
 
   ros::Duration max_storage_time_;
-
+  ros::Time last_repeated_warn_time_;
```
peci1, line 121: This most probably wouldn't make it into Noetic. It changes the ABI of existing code, which would require a total rebuild of the whole Noetic buildfarm.

I suggest moving this completely to the .cpp . You can make this a static variable in the .cpp file, which would preserve ABI. It could also simplify the implementation.

If you no longer feel the urge to get this fix merged, ping @klaxalk , he told he could find some time to dedicate to this fix ;)

peci1 (changes requested):

peci1:
Alternatively, this problem could be overcome by adding dependency on rosconsole_bridge and putting `REGISTER_ROSCONSOLE_BRIDGE` in some .cpp file that's always loaded. This would redirect all buffer messages to standard rosconsole, where the users could deal with the messages as they like.

However, I don't like this alternative too much - it would change the logging behavior someone might be relying on. Also, it would not allow to selectively discard only the TF_REPEATED_DATA messages.

sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one. It changes ABI, and ABI breaking changes cannot be merged into released ROS distros.

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160)

PR #519, opened 2021-09-20
Add transform broadcaster constructors which can take a node handle as an argument.

malban:
When remapping `/tf` or `/tf_static` for a node or nodelet in roslaunch, the topics are only remapped correctly when the broadcaster is part of a node.  If the broadcaster is part of a nodelet, the topics are not remapped because the broadcasters are publishing on the manager's node handle instead of the nodelet's node handle.

This MR adds new constructors for the transform broadcaster and static transform broadcaster that take in a user provided node handle.   This allows a nodelet to pass it's node handle to the broadcaster constructors.

### DIFF
--- a/tf2_ros/include/tf2_ros/static_transform_broadcaster.h
+++ b/tf2_ros/include/tf2_ros/static_transform_broadcaster.h
@@ -52,6 +52,9 @@ public:
   /** \brief Constructor (needs a ros::Node reference) */
   StaticTransformBroadcaster();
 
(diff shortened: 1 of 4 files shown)


tfoote:
Thanks in general this would be a great extension to support your use case. However as we've already released this into melodic and noetic I don't want to extend the API and break the ABI for existing users. And there's no targets for development beyond melodic or noetic. Thus despite this being a useful extension to the API I don't have a good place to merge it. 

New development is focused on the ROS 2 port at https://github.com/ros2/geometry2

PR #522, opened 2021-11-03
Restore time difference order so future extrapolation exceptions show positive seconds into future

lucasw:
There was a manual merge 52bff88ecad8821802ad5e04dcb711b582c39f5a as a part of #470  that caused this regression, resulting in negative seconds into the future:

```
tf2.ExtrapolationException: Lookup would require extrapolation -0.041266203s into the future.
```

If the t0 - t1 ordering is confusing could alternatively do `-tdiff.to_sec()`.


tfoote (approved):
Yeah, this is the correct behavior.

PR #527, opened 2022-02-17
Updating the documentation to current constructor for a MessageFilter

tassos:
I think there is an outdated example in the documentation for the MessageFilters. Please correct me if I am wrong.


tfoote (approved):
Thanks for catching this and submitting a fix.

tassos:
My pleasure :) Thanks for merging this!

PR #528, opened 2022-02-18
Switch to new boost/bind/bind.hpp

jspricke:


tfoote:
Can you provide some context as to why we should switch to it? Clearly it's more explicit. From your title it appears to be a new syntax. What platforms will this enable and how far backwards is it compatible?

jspricke:
The old syntax is deprecated and gives a warning with recent Boost versions (didn't check with version). The new syntax is compatible with at least Ubuntu 18. Actually nowadays there is also `std::bind` and lambda functions but this works best with all platforms and versions.

tfoote (approved):
Wow digging in I found out that this was initially updated in 2009: https://github.com/boostorg/bind/commit/8f507b9aeca643ca78e6a712b6d300720627c0ed 

And the clear recommendation was added 2 years ago: https://github.com/boostorg/bind/commit/5c21b6a5251523453ef3d02cc7e44dbfb31275eb

https://github.com/boostorg/bind/commit/2797f0dc33caaae126a416bf613bd11267ba3353

Doing this makes sense and it's actually using the same implementation just without the redirect and with more explicit scoping so it shouldn't break anyone. The only risk would be downstream packages which are relying on the namespace leakage of the global declarations. But they can fix that by including the old bind.hpp which is the syntax that they're actually using.

PR #530, opened 2022-03-22
Removed print statements from buffer interface

jlunenburg:


MatthijsBurgh (approved):

MatthijsBurgh:
@tfoote friendly ping ;)

tfoote (approved):
Thanks yeah, these are unnecessary and can be cleaned up. 

The documentation reads that it was potentially used for testing more than expected to be used as the primary method. But I think it does appear to be the primary usage.

PR #532, opened 2022-04-18
Fix dead loop in message filter

kingsimba:
The bug is introduced in https://github.com/ros/geometry2/pull/461

`std::remove_if()` will only put elements to the back of the vector. One must call erase to actually remove them.


tfoote (approved):
Wow, thanks for catching that. It definitely can lead to unfortunate memory growth. 

There's a much better function [`erase_if`](https://en.cppreference.com/w/cpp/container/vector/erase2) which is what I had thought it was doing. But that's not available until `c++20` so we can't use it yet.

PR #534, opened 2022-04-19
remove useless method with misleading doc

wvmcastro:
The overload 

`void tf2::fromMsg ( const geometry_msgs::TransformStamped & msg, geometry_msgs::TransformStamped &)`

seems to be useless since it consists of a call to the copy constructor only. Nonetheless, its documentation is wrong.


tfoote (changes requested):
Yeah the docs are wrong, but this seemingly useless method is required for the templated conversion logic to have coverage.

wvmcastro:
Ok, i will edit just the doc part and request another review, tks

tfoote (approved):
Thanks for the update.

PR #535, opened 2022-05-06
Bugfix: SetEuler/SetRPY implementations were differing

guilyx:
Hi there, small bug fix along with a unit test.

SetEuler takes yaw, pitch, roll as parameters in that order, SetRPY takes roll, pitch, yaw in that order. They should produce the same Quaternion when passed the same yaw/pitch/roll in the right respective order, but it doesn't.
The explanation is simple: SetEuler and SetRPY don't have the same implementation even though they should (the order changes but the way to calculate the Quaternions stay the same).

In the current implementation, the unit test I added was failing, now it's not.


tfoote (changes requested):
Thanks for looking at this however the unit tests you've written are not correct. Euler angles and fixed axis angles are not the same and are not expected to produce the same results given the same 3 angle values. That's why there are different methods exposed in the API with different implementations for each context.

PR #144, opened 2016-01-14
Solve a bug that causes a deadlock in MessageFilter

afakihcpr:
The deadlock happens when the following two things happen at the same instant:
- MessageFilter trying to erase a message from the queue, locks the messages_mutex_ then calls BufferCore::cancelTransformableRequest where it needs the transformable_requests_mutex_
- BufferCore upon receiving a TF callback, locks the transformable_requests_mutex_ in testTransformableRequests and goes through the list of requests where one of the callback it calls is MessageFilter::Transformable that attempts to acquire a unique_lock on message_mutex_

related to https://github.com/ros/geometry_experimental/issues/91

@tfoote @jasonimercer @efernandez @ayrton04

Backtrace:

``` bash
Thread 6 (Thread 0x7fffedc55700 (LWP 28549)):
#0  pthread_cond_wait@@GLIBC_2.3.2 ()
    at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00007ffff7a4ae4f in boost::condition_variable::wait (this=0x62e0f0, 
    m=...) at /usr/include/boost/thread/pthread/condition_variable.hpp:73
#2  0x00007ffff7a4b924 in boost::shared_mutex::unlock_upgrade_and_lock (
    this=0x62e010) at /usr/include/boost/thread/pthread/shared_mutex.hpp:503
#3  0x00007ffff7ad0272 in boost::unique_lock<boost::shared_mutex>::unique_lock
    (this=0x7fffedc54078, other=...)
    at /usr/include/boost/thread/lock_types.hpp:1046
#4  0x00007ffff7ac23fd in boost::upgrade_to_unique_lock<boost::shared_mutex>::upgrade_to_unique_lock (this=0x7fffedc54070, m_=...)
    at /usr/include/boost/thread/lock_types.hpp:1063
#5  0x00007ffff7ab261d in tf2_ros::MessageFilter<geometry_msgs::TwistWithCovarianceStamped_<std::allocator<void> > >::transformable (this=0x62df50, 
    request_handle=1358, target_frame=..., source_frame=..., time=..., 
    result=tf2::TransformAvailable)
    at /home/afakih/workspaces/rl_ws/src/geometry_experimental/tf2_ros/include/tf2_ros/message_filter.h:542
[...]

PR #538, opened 2022-07-01
Fix race condition in MessageFilter

rhaschke:
A bug reported in https://github.com/ros-visualization/rviz/issues/1753 pointed to race condition(s) in `tf2_ros::MessageFilter`.
I implemented a stress test to analyze the problem in detail. There were two issues:
1. A race condition routing back to #91, #101, #144 
   As pointed out in #144 the deadlock arises due to an inversion of locking order:
    - `MessageFilter::add()`, when removing the oldest message, locks:
      - `MessageFilter::messages_mutex_`
      - `BufferCore::transformable_requests_mutex_` (via `cancelTransformableRequest()`)
    - `BufferCore::testTransformableRequests()` locks:
      - `transformable_requests_mutex_`
      - `MessageFilter::message_mutex_` (via `MessageFilter::transformable()` callback)
2. A race condition with `CBQueueCallback`
   When `MessageFilter::signalMessage()` is called asynchronously via a ROS callback queue, this call might still be in progress (holding `Signal1::mutex_`), while another thread might attempt to remove the MessageFilter.
This results in a failed assertion as the MessageFilter's destructor attempts to destroy a locked mutex.

PR #144 attempted to resolve the first issue by postponing callback calls. However, this has the drawback of using references to `TransformableRequests` outside the protection by `transformable_requests_mutex_`. Thus, these requests might get deleted by another thread, resulting in a segfault as revealed by the stress test.
Here the deadlock is resolved in `MessageFilter::add`, canceling the requests outside the `messages_mutex_` protection.

To resolve the second race condition, another mutex is introduced in MessageFilter, to ensure that `CBQueueCallback::call()` is finished before destroying the MessageFilter.

These two changes make the stress test pass.
@tfoote, please note that this is a PR against Melodic. I will prepare a port for Noetic as well.

### DIFF
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -1462,7 +1462,7 @@ void BufferCore::testTransformableRequests()
     }
   }
 
-  // unlock before allowing possible user callbacks to avoid potential detadlock (#91)
+  // unlock before allowing possible user callbacks to avoid potential deadlock (#91)
   lock.unlock();
 
   // Backwards compatability callback for tf
--- a/tf2/src/buffer_core.cpp
+++ b/tf2/src/buffer_core.cpp
@@ -37,7 +37,6 @@
 #include <assert.h>
 #include <console_bridge/console.h>
 #include "tf2/LinearMath/Transform.h"
-#include <boost/foreach.hpp>
 
 namespace tf2
 {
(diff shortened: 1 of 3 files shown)


rhaschke:
Looks like there is still a deadlock :disappointed:

rhaschke:
> Looks like there is still a deadlock

Resolved as well. 
But I still see occassionally a race condition when using a ROS callback queue to signal messages, causing the assertion failure of issue (2).

rhaschke:
I'm afraid the remaining race condition cannot be fixed. Rather, the ROS callback queue mechanism should be avoided.

While `MessageFilter::clear()` removes pending messages from the callback queue, there might be still callbacks active when attempting to remove the MessageFilter. Depending on what goes first, either the already destroyed mutex is tried to lock
or a locked mutex is tried to destroy, both resulting in an assertion failure.

ahcorde:
This PR targets `melodic` which is [EOL](http://wiki.ros.org/Distributions), closing this PR. There is already another PR targeting `noetic` https://github.com/ros/geometry2/pull/539.

rhaschke:
I'm fine that you closed this PR, @ahcorde. But it is a pity that this PR was lying around for 19 months w/o any feedback and now being closed due to EOL reasons. I hope the Noetic port (#539) will be considered at some point :wink:

PR #539, opened 2022-07-02
Fix race conditions in MessageFilter

rhaschke:
This is a port of #538 to Noetic.


tf2/src/buffer_core.cpp
```diff
@@ -1455,12 +1449,8 @@ void BufferCore::testTransformableRequests()
         M_TransformableCallback::iterator it = transformable_callbacks_.find(req.cb_handle);
         if (it != transformable_callbacks_.end())
         {
-          transformables.push_back(boost::make_tuple(boost::ref(it->second),
```
rhaschke, line 1459: The objects referenced here might get destroyed before they are used below, resulting in a segfault.

tf2_ros/include/tf2_ros/message_filter.h
```diff
@@ -215,14 +215,15 @@ class MessageFilter : public MessageFilterBase, public message_filters::SimpleFi
   ~MessageFilter()
   {
     message_connection_.disconnect();
-
     MessageFilter::clear();
+    bc_.removeTransformableCallback(callback_handle_);
```
rhaschke, line 219: The `callback_handle_` should be deregistered with `BufferCore`.

tf2/src/buffer_core.cpp
tfoote: I need to look a little deeper that this isn't going to restore the potential deadlock of #91

tfoote (commented):
Overall this looks like a good fix for the MessageFilter destruction which was not clearly accounted for. And the unit tests for it are very helpful. I want to give the feedback here but I want to read/analyze a little bit deeper in the BufferCore changes, but I don't have time right now to do so.

tf2/src/buffer_core.cpp
rhaschke: Sure. Fortunately, unit tests nicely cover this deadlock. While https://github.com/ros/geometry2/pull/539/commits/23e6295ebb14279892fd55b7141d6f50b7ed0f25 restores the deadlock (as expected), resulting in a `RLTestTimeoutException`, https://github.com/ros/geometry2/pull/539/commits/9fdf5474907bf1b9d0983ebdeb8b25c93a487960 fixes it again using the method described in #538.

rhaschke:
Ping @tfoote.

rhaschke:
@tfoote, would be great if you could have the deeper look you promised nearly a year ago...

tf2_ros/include/tf2_ros/message_filter.h
```diff
@@ -668,7 +670,8 @@ class MessageFilter : public MessageFilterBase, public message_filters::SimpleFi
   V_string target_frames_; ///< The frames we need to be able to transform to before a message is ready
   std::string target_frames_string_;
   boost::mutex target_frames_mutex_; ///< A mutex to protect access to the target_frames_ list and target_frames_string.
-  uint32_t queue_size_; ///< The maximum number of messages we queue up
+  boost::shared_mutex cbqueue_mutex_; ///< A mutex protecting calls from callback queues
```
sloretz, line 673: It looks like this changes the size of `CBQueueCallback`. I'm not totally sure if the ABI of `CBQueueCallback` matters, but I don't have time to look deeper at this. I suspect this PR is unlikely to make it in before Noetic eol.

That's probably pretty disappointing given #538 being closed for Melodic going EOL, but not much we can do about it now.

tfoote: Thanks for flagging that @sloretz I think it's ok. It looks like a private struct inside the templated class. It's only directly used to push the data which is dynamically allocated into the queue on either `messageReady` or `messageDropped` and then the rest of the class only accesses the private queue of the `MessageFIlter` class member pointer `callback_queue_` which is then only leverages the generic interface. 

Because it's all in private elements and compiled into the implementation by the template I believe that this should not break ABI as there's no external interface accessible to this datatype, and any instantiation of this template should not be interacting with other instances in other compile units.

tfoote (approved):
Thanks for your patience. I've read this through and it has a clean alternative solution to the #91 issues as well as then resolving the other identified race condition. 

I will sync with @sloretz about risk levels of merging and to double check my ABI analysis.

tf2_ros/include/tf2_ros/message_filter.h
sloretz: Thanks for reviewing Tully! I double checked, and I think you're correct. Merging.

sloretz:
Thank you for the PR!

PR #544, opened 2022-11-17
Fix(tf2_py) potential memory leak

MatthijsBurgh:
I think the chance of `pargs = Py_BuildValue("()");`  resulting in a `NULL` is very small. Though it is checked and so when entering this `if` `plcass` is not `NULL` and should have his ref counter decreased.


tfoote (approved):
Yeah, this is the right thing to do when the logic exits early. I too don't think it's likely for anyone to hit this. I'm not familiar with known use cases ever hitting this branch/error condition or how to even trigger it except maybe memory exhaustion which will have much bigger issues.

PR #545, opened 2022-11-29
Fix TypeError when using --cache_time argument

tik0:
Running `rosrun tf2_tools echo.py --cache_time 1.0 source_frame target_frame` causes `TypeError: time must have a to_sec method, e.g. rospy.Time or rospy.Duration`, because the argument is not properly converted. The fix handles the conversion, if the argument was given.

### DIFF
--- a/tf2_tools/scripts/echo.py
+++ b/tf2_tools/scripts/echo.py
@@ -139,7 +139,7 @@ def _euler_from_quaternion_msg(quaternion):
 
 class Echo():
     def __init__(self, args):
-        self.tf_buffer = tf2_ros.Buffer(cache_time=args.cache_time)
+        self.tf_buffer = tf2_ros.Buffer(cache_time=rospy.Duration(args.cache_time) if args.cache_time else None)
         self.tf_listener = tf2_ros.TransformListener(self.tf_buffer)
         self.args = args
 


tfoote (commented):
Thanks for catching this it's a good fix. 
This is the minimum fix and looks fine. 

Looking at the if statement I think it would be better to actually pass in the default cache time value to the arguments for better visibility. 

It's in c++ here as `tf2::BufferCore::DEFAULT_CACHE_TIME`
We likely need to expose it into the python library. https://github.com/ros/geometry2/blob/noetic-devel/tf2_py/src/tf2_py.cpp and then reference that in the command line usage as [a default here](https://github.com/ros/geometry2/blob/c92cd9a9073c255388599495d2f86799ecafa1bb/tf2_tools/scripts/echo.py#L231).

sloretz:
@tfoote it looks like passing `None` results in DEFAULT_CACHE_TIME being used:

https://github.com/ros/geometry2/blob/44a604df70c1fd1b596a0e5e791c4c549e5c711c/tf2_ros/src/tf2_ros/buffer.py#L58-L62

https://github.com/ros/geometry2/blob/44a604df70c1fd1b596a0e5e791c4c549e5c711c/tf2_py/src/tf2_py.cpp#L181-L186

so this LGTM. Any objections to merging this?

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #547, opened 2023-01-20
add missing link to librt

lianghongzhuo:
compile tf2_ros get link error
```
undefined reference to 'clock_gettime'
```
solution is add rt in the link
reference: https://stackoverflow.com/questions/2418157/c-error-undefined-reference-to-clock-gettime-and-clock-settime


lianghongzhuo:
strange, this error not exist anymore, so close this pr.

PR #553, opened 2023-08-01
Use only find_package to find bullet.

mikepurvis:
The prior logic was failing on NixOS 23.05; using CMake always should now be safe, as even Bullet 2.88 in Ubuntu Focal contains the CMake module:

```
/usr/lib/x86_64-linux-gnu/cmake/bullet/BulletConfig-float64.cmake
/usr/lib/x86_64-linux-gnu/cmake/bullet/BulletConfig.cmake
/usr/lib/x86_64-linux-gnu/cmake/bullet/UseBullet.cmake
```

https://packages.ubuntu.com/focal/amd64/libbullet-dev/filelist

### DIFF
--- a/tf2_bullet/CMakeLists.txt
+++ b/tf2_bullet/CMakeLists.txt
@@ -1,24 +1,13 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(tf2_bullet)
 
-find_package(PkgConfig REQUIRED)
-
-set(bullet_FOUND 0)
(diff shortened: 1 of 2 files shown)


tfoote (approved):
Thanks for the simplification. I also checked it's this way on buster too(which was an official target): https://packages.debian.org/buster/amd64/libbullet-dev/filelist

And the Windows was already relying on the CMake approach too.

PR #554, opened 2023-09-05
add torque due to force offset

v4hn:
adapted patch by @pac48 in https://github.com/ros2/geometry2/pull/538 .
Thank you for the original patch.
Impressive to see how this was undetected in the sources for four years...

@destogl


tfoote (approved):
Thanks for backporting this here too.

Issue #541, opened 2022-09-15
dotransform TransformStamped datatype not process child_frame_id


darkchooselight:
the code do not process child_frame_id transform. look below:
tf2_geometry_msgs/tf2_geometry_msgs.h

```
template <>
inline
void doTransform(const geometry_msgs::TransformStamped& t_in, geometry_msgs::TransformStamped& t_out, const geometry_msgs::TransformStamped& transform)
  {
    tf2::Transform input;
    fromMsg(t_in.transform, input);

    tf2::Transform t;
    fromMsg(transform.transform, t);
    tf2::Transform v_out = t * input;

    t_out.transform = toMsg(v_out);
    t_out.header.stamp = transform.header.stamp;
    t_out.header.frame_id = transform.header.frame_id;
  }
```

tfoote:
That does look like it's missing the logic to transfer the `child_frame_id` if t_in and t_out aren't the same.

tfoote:
I was reviewing the PR #555 proposed to resolve this and I think this behavior is actually correct. See my logic there.

PR #555, opened 2023-09-17
#541 [BUG] child_frame_id not processed

oss-vineet:


tfoote (commented):
Thanks for putting this in. One review though I think the current behavior is correct. aTb * bTc => aTc. As such only one of the coordinate frame headers should change.

oss-vineet:
@tfoote got it, thanks! This didn't take much effort, so not a problem. Let me close this then.

PR #558, opened 2024-01-08
Add missing #include to buffer_core.cpp

vslashg:
This file uses `boost::tuple` without including the associated header.  Changes to Boost 1.84 cause this name to be no loner visible from this .cpp file.


tfoote (approved):
Thanks, for fixing the missed indirect dependency.

PR #559, opened 2024-01-26
Fixed error message when fixed_frame is not found

peci1:
Looks like copy&paste error. Should only alter the warning messages, not the behavior of the library.


tfoote (approved):
Yeah, that looks like a typo.

We should make sure it's fixed in the ROS 2 for as well.

PR #561, opened 2024-02-03
tf2_eigen: Deprecate methods using Affine3d (Noetic)

peci1:
Noetic continuation of #371.

I've also silenced the warnings generated by the unit test. The pragmas with `GCC` are also understood by clang, so these two compilers should be covered.


ahcorde (approved):
@tfoote ?

sloretz:
Thank you for the PR!

ROS Noetic is a released distro, so we can't introduce deprecation warnings into it now.

[ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160)

Issue #486, opened 2020-11-25
tf2_py build failure with catkin symlink install


dgoel:
`tf2_py` project errors out when building with symlink install option for `catkin` with `catkin-tools` (or `colcon`). This does not happen if symlink option is *not* enabled.

The main cause is that `catkin` does not handle cmake generator expression in the install functions for cmake. Nevertheless, proposing a fix here.

## Version(s) affected

I have only tested with `melodic-devel` but this should also affect `noetic-devel`.

## How to reproduce

With ros melodic distro:

```shell
# clone
$ mkdir -p ros_ws/src
$ cd ros_ws
$ git clone https://github.com/ros/geometry2 src

# build
$ catkin init
$ catkin config --install --cmake-args -DCATKIN_SYMLINK_INSTALL:BOOL=ON
$ catkin build
```

Build error:
```
CMake Error at catkin_symlink_install/catkin_symlink_install.cmake:157 (message):
  catkin_symlink_install_files() can't find
  '/home/user/ros_ws/src/tf2_py/$<TARGET_FILE:tf2_py>'
Call Stack (most recent call first):
  catkin_symlink_install/catkin_symlink_install.cmake:343 (catkin_symlink_install_files)
  cmake_install.cmake:41 (include)


make: *** [install] Error 1
```

## Proposed fix

This can be fixed by not using cmake generator expression in the install as follows:

```patch
$ git diff
diff --git a/tf2_py/CMakeLists.txt b/tf2_py/CMakeLists.txt

-install(FILES $<TARGET_FILE:tf2_py>
+install(TARGETS tf2_py
     DESTINATION ${CATKIN_PACKAGE_PYTHON_DESTINATION}
 )
```

I am happy to open a PR if this is acceptable.

mqcmd196:
+1

mqcmd196:
This is fixed via https://github.com/ros/geometry2/pull/567 . We can close this.

PR #567, opened 2024-09-05
[ROS-O] tf2_py build failure with catkin symlink install

v4hn:
@mqcmd196 contributed [a PR to the ROS-O repository](https://github.com/ros-o/geometry2/pull/1), but the upstream PR did not exist yet.

Addresses https://github.com/ros/geometry2/issues/486 with the proposed fix.


sloretz (approved):
Thank you for the PR!

